PR #22, opened 2026-07-19
Type-guard narrowing for rlang standalone checkers and stopifnot/assert_that

### LINKED ISSUE #21: Support type guards like rlang's types-check standalone
This is a very cool project!

I'm curious if you have plans to support type guard functions? For example, rlang provides [a number of functions](https://github.com/r-lib/rlang/blob/main/R/standalone-types-check.R) that don't directly return a value but enforce type constraints. One complication is that these functions are often inlined directly into package code with `usethis::use_standalone()`.

For example, this code would produce a length > 1 error

```r
if (choice == "foo") {
  # action...
}
```

but this might be made safe by a `check_string()` enforcing that `choice` is a length-1 string

```r
check_string(choice)

if (choice == "foo") {
  # action...
}
```

---

This might be a bit of stretch given the general constraints of working with a dynamically typed language without annotations, but I tried `ry` on a few packages I work with and immediately ran into a few of these kinds of false positives. Either way, thanks for working on this, I've already found it useful!

sims1253:
Closes #21.

Bare calls to the inlined `standalone-types-check.R` family (`check_string()`, `check_bool()`, `check_number_whole()`, ...) are now understood as type guards: execution past the call proves the asserted type of the first argument, so downstream diagnostics (e.g. RY002 on `if (choice == "foo")` after `check_string(choice)`) no longer fire.

## Design

- Extends the existing assertion-narrowing mechanism (`assertion_call_target`, previously covering admiral's `assert_*_scalar` family) with a `standalone_check_target` table: scalar targets for `check_bool`/`check_string`/`check_name`/`check_number_*`/`check_function`/`check_closure`, class targets for `check_symbol`/`check_arg`/`check_call`/`check_formula`/`check_environment`/`check_data_frame`, unknown-length vector targets for `check_character`/`check_logical`. `check_required` proves nothing about type and is excluded.
- **Fingerprint gating**: these names are conventional, not reserved. When a same-named user definition exists, narrowing applies only if its formals include `arg` and `call` — the signature every standalone checker has (`function(x, ..., arg = caller_arg(x), call = caller_env())`). A non-fingerprinted collision is inferred as an ordinary call. When no definition is in view (single-file checks), the name is trusted, which is the FP-suppressing direction.
- **`allow_null` / `allow_na`**: absent or literal `FALSE` keeps the full claim; literal `TRUE` or a non-literal value weakens the target by joining NULL (resp. scalar logical for a bare `NA`). The length-1 claim survives the weakening, which is what matters for the condition-length FPs.
- **`stopifnot()` / `assert_that()`**: each argument runs through the existing if-condition narrowing extraction and applies its positive (then-branch) refinement to the continuation scope — correct for all narrowing kinds since execution continues exactly when every argument is TRUE. `assert_that`'s `msg =` argument is skipped. Arguments are still inferred normally first, so diagnostics inside them are preserved.

The inverse diagnostic (`check_string(x)` on a known-incompatible `x` always throws — flag the guard, treat the continuation as unreachable) is deliberately left for a follow-up.

## Validation

- 8 new unit tests: name-trust, fingerprint-accept, collision-reject, `allow_null` weakening, `check_number_whole` scalar union, `stopifnot`/`assert_that` positive narrowing, namespaced `assert_that` with `msg` exclusion.
- Full `ry-checker` suite green (385 unit tests + corpus/oracle/project/real-world/vendor suites); vendor snapshots unchanged.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` clean.
- End-to-end repro: known length-2 vector + `check_string()` + `if (==)` is silent; the same code without the guard still fires RY002.

### DIFF
--- a/crates/ry-checker/src/infer/call.rs
+++ b/crates/ry-checker/src/infer/call.rs
@@ -652,6 +652,48 @@ impl Checker {
             }
             return RType::new(Mode::Null, Length::Zero);
         }
+        if !name.contains("::")
+            && let Some(mut target) = standalone_check_target(&lookup_name)
+            && user_function.as_ref().is_none_or(|function| {
+                ["arg", "call"].into_iter().all(|required| {
+                    function
+                        .params
+                        .iter()
+                        .any(|parameter| parameter.name == required)
+                })
+            })
+            && let Some(Expr::Ident { name: var, .. }) = args.first().map(|a| &a.value)
+        {
+            // A non-literal opt-in is treated like TRUE: weakening the fact
+            // avoids excluding a value the checker may accept at runtime.
+            if args.iter().any(|argument| {
+                argument.name.as_deref() == Some("allow_null")
+                    && !matches!(argument.value, Expr::Logical(false, _))
+            }) {
+                target = target.join(RType::new(Mode::Null, Length::Zero));
+            }
+            if args.iter().any(|argument| {
+                argument.name.as_deref() == Some("allow_na")
+                    && !matches!(argument.value, Expr::Logical(false, _))
+            }) {
+                target = target.join(RType::scalar(Mode::Logical));
+            }
+            scope.insert(var.clone(), target);
+            return RType::new(Mode::Null, Length::Zero);
+        }
+
+        let assertion_predicates =
+            name == "stopifnot" || name == "assert_that" || name == "assertthat::assert_that";
+        if assertion_predicates {
+            for argument in args {
+                if name.ends_with("assert_that") && argument.name.as_deref() == Some("msg") {
+                    continue;
+                }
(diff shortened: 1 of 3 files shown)


sims1253:
@gadenbuie what do you think of this?

gadenbuie:
Nice! My initial impression is it's defensively brittle; probably the best we can get without annotations 😉 Thanks for the quick response!

sims1253:
Yea, I have had thoughts on type annotations but that is such a big thing that should probably include people that have more experience with this kind of tooling than me, who is essentially just yelling at my bot :D
The current beauty of ry is that it has benefits without requiring any additional work from maintainers.

Issue #81, opened 2026-08-12
LSP: close/reopen publishes no diagnostics (convergence property fails on main)


sims1253:
## Summary

`w10_session_converges_to_fresh_server` fails on `main`/`plans-37-39-release-prep`.
After a close/reopen, the live server publishes **no diagnostics** for the
reopened document while a fresh server publishes four.

This is the defect `d8bedcc` ("Fix convergence bug: force full re-collection
after file changes") claimed to fix. It does not fix it — the failure reproduces
with `force_full_recollection` in place.

## Reproduction

```
cargo test -p ry-lsp --test session w10_session_converges
```

Minimal failing sequence:

```
Open  { file: 1, source: BmpDiagnostic }
Open  { file: 1, source: AsciiDiagnostic }
Close { file: 1 }
IncrementalEdit { file: 0, source: AsciiDiagnostic }
Open  { file: 1, source: AstralDiagnostic }     <- fails here (step 4)
...
```

Observed at step 4 for `b.R`:

```
live:  []
fresh: [
  RY000 "syntax error: unparseable region (recovered tree may be unreliable)" 0:0-0:2,
  RY000 "syntax error: unparseable region (recovered tree may be unreliable)" 0:0-0:5,
  RY090 "unknown argument `xx` to `length`; did you mean `x`?" 0:13-0:20,
  RY091 "missing required argument `x` in call to `length`" 0:6-0:21,
]
```

## Why it was believed fixed

Two masking effects:

1. **No seed was ever committed.** `crates/ry-lsp/tests/session.proptest-regressions`
   contained only its header, so no run ever replayed a known-bad sequence.
   (Proptest does write the seed on failure, despite emitting
   `FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs`
   — the write succeeds. The seed had simply never been committed.) A seed is now
   recorded: `cc fe14e0af97a076ae42ae0bbe16d9234da5e8b99c91947800e59439bd56e7abe9`.
   With it present the failure is deterministic and reproduces in ~3.5s.
2. **Low case count.** `proptest_config` hardcodes `cases: 8`, so runs sample a
   small number of sequences and frequently miss it.

Together these made the failure look intermittent and "pre-existing but
unreproducible". `docs/architecture/p38-w12-acceptance.md` records
"963 workspace tests pass (1 pre-existing convergence failure)" — that failure
is this one, and it is still present.

## Evidence that `force_full_recollection` is not the fix

[...]

PR #95, opened 2026-08-14
Test theater sweep (#82): fix or delete assertions that could not fail

### LINKED ISSUE #82: Test theater: assertions that pass regardless of behaviour
Tests that pass regardless of the behaviour they claim to check. Found during
the PR #79 review round. Each is cheap individually; together they are why the
suite reported green while real defects shipped.

Marked **[verified]** where I read the code and confirmed, **[reported]** where
it comes from the review bot and has not been independently checked.

### Assertions that cannot fail

- **[verified]** `crates/ry-analysis/src/check.rs:102` —
  `check_project_with_workspace_context` builds a `WorkspaceContext` with
  `dplyr` attached, then asserts only `assert!(!output.diagnostics.is_empty())`.
  It never compares against the same input *without* the workspace context, so
  it would pass identically if the context were ignored entirely. Fix: assert
  the attached-package run differs from the empty-workspace run in the specific
  expected way.

- **[verified]** `crates/ry-lsp/tests/p38_feature_diff.rs:187` —
  `assert!(target_uris.iter().all(|uri| *uri != a_uri), "definition should
  resolve to the local b.R helper...")` is vacuously true when `target_uris` is
  empty, i.e. when go-to-definition returns nothing at all. Fix: assert
  non-empty first, then assert the target.

- **[reported]** `crates/ry-analysis/src/symbols.rs:435` — the test does not
  verify the behaviour its comment claims.

### Properties that do not vary what they claim to vary

- **[verified]** `crates/ry-analysis/tests/property.rs:129` —
  `content: format!("v{}", n)` uses the loop *bound*, not the loop index, so
  every iteration writes byte-identical content. The property intends distinct
  content per revision.

- **[reported]** `crates/ry-analysis/tests/property.rs:52` — the live-vs-fresh
  property does not reconstruct the same inputs for the two sides, so the
  equality it asserts is weaker than intended.

### Harness weaknesses

- **[reported]** `crates/ry-lsp/tests/p38_feature_diff.rs:44` — fixed sleep for
  indexing instead of a bounded poll on observable state. The project's own
  testing principle is "wait on observable state or protocol events, never
  arbitrary sleeps".

- **[reported]** `crates/ry-lsp/tests/w8_session.rs:292` — two validity rules
  narrow the generated sequences enough to reduce the coverage the test was
  added to provide.

- **[reported]** `crates/ry-lsp/tests/w8_session.rs:593` — the counter name and
  the reported message disagree with what is actually counted.

### Related

`cases: 8` is hardcoded in the `w10_session_converges_to_fresh_server` proptest
config (`crates/ry-lsp/tests/session.rs:876`), which is a large part of why the
convergence defect in #81 went unreproduced for so long. Worth raising here or
in #81.

### Acceptance

For each item: the test must fail when the behaviour it names is broken.
Demonstrate that by breaking the behaviour locally and observing the failure,
not by inspection alone.

sims1253:
Closes #82.

Every [verified]/[reported] item was checked against current `main`; each was either fixed to an assertion that provably fails, deleted, or documented down to what it actually checks — per the issue's acceptance criterion, demonstrated by mutating production code and observing failures, not inspection.

## Disposition per item

1. **[verified] `ry-analysis/src/check.rs` — workspace-context test.** Was: attach dplyr, assert diagnostics non-empty (passes with the context ignored entirely). Now: same `rlang`-narrowing source run with and without the context, asserting the RY070 flip in both directions. *Mutation: forcing `check_project` onto an empty workspace fails the with-rlang run (RY070 1 != 0).*
2. **[verified] `p38_feature_diff.rs:187` — vacuous assertion.** Stale: the file was deleted in #79. Pruned from the issue.
3. **[reported] `ry-analysis/src/symbols.rs` — resolution claim.** Verified real: `SymbolIndex` has no position-resolution API, so the comment described behavior the API cannot express while the assertion checked a start offset any merge produces. Now asserts both per-file definitions survive the merge and b.R's call stays indexed as a reference. *Mutation: disabling `add_reference` fails the references assertion (0 != 1).*
4. **[verified] `ry-analysis/tests/property.rs` — loop bound as content.** `format!("v{}", n)` → `format!("v{i}")`, so successive applies actually change content. *Mutation: a host that skips the revision bump when a write changes existing content fails the fixed test (shrinks to n=2); the old body passes under the same mutation — the old test could not see that behavior.*
5. **[reported] `property.rs` — live-vs-fresh inputs.** The host's public API (`apply`, `revision`, `file_content`, `all_files`, `roots`, `snapshot`, `config`) exposes no open-file queries, so the fresh side cannot reconstruct open/closed state or versions. Per the no-new-API rule, the doc comment and property description now claim only final-contents equivalence — which is what the property always compared. No mutation applies to a documentation change.
6. **[reported] `p38_feature_diff.rs:44` — fixed sleep.** Stale: file deleted in #79. Pruned.
7. **[reported] `w8_session.rs` — validity funnel.** Verified real: edits-require-open plus free-slot conjunctions invalidate ~9/22 of the alphabet at sequence start, and corrections always retargeted the lowest slot. The validity rules themselves are protocol correctness and stay; corrections now rotate across candidate slots via model state (deterministic — the seed lanes replay identically). **Honest numbers:** on the 16-seed PR lane, distinct edit-target slots are 3 vs 3 — the lane is underpowered for this metric (only 3 seeds produce any corrected edit op, exactly one each); across all corrected op types, rotation does move targets (e.g. seed `0x0fed…` hits slots [1,2] rotated vs [0,1] always-first). No coverage regressed; no convergence failures surfaced.
8. **[reported] `w8_session.rs` — counter/message mismatch.** Verified real: counted ops failing `is_valid` (a superset of both "corrected" and "skipped") while reporting them as "corrected", with a comment claiming it "should remain zero" (it almost never was). Deleted — counter, increment, comment, and eprintln. Not replaced.

The issue's Related note (`cases: 8` in the w10 config) was resolved in #81/#93: the failure it hid was a harness race, the seeds are committed and replay every run, and a deterministic regression test plus a one-off 256-case run during #94 gate the property.

## Gates

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all green. Mutation artifacts (a proptest regressions entry pinned to the deliberately broken case) were removed; the tree contains only the reviewed changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type-narrowing diagnostics when the required package is available or missing.
  - Improved language server session recovery by rotating through eligible file targets, helping sessions converge more reliably.
  - Corrected revision tracking behavior across repeated file updates.
  - Improved duplicate-symbol indexing validation for definitions and references.

- **Tests**
  - Strengthened analysis and language server coverage.
  - Removed misleading stale-coverage and skipped-operation reporting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,16 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Tests that could not fail were fixed or deleted (#82): the ry-analysis
+  workspace-context test now asserts the rlang/RY070 flip in both
+  directions (check.rs), the same-name symbol test asserts reference
+  indexing instead of an unobservable resolution claim (symbols.rs), the
+  property tests write distinct per-apply content and claim only
+  final-contents equivalence (tests/property.rs), and the w8 session
+  model rotates corrected-operation targets across candidate slots with
+  the stale coverage eprintln (which miscounted invalid ops as
+  corrections and claimed it should stay zero) removed
+  (ry-lsp/tests/w8_session.rs).
 - `Checker::check` now resets the function table and return slots before each
   run, so a reused single-file `Checker` no longer leaks inference state
   (functions, known-vars) from a previous file into the current check. This
--- a/crates/ry-analysis/src/check.rs
+++ b/crates/ry-analysis/src/check.rs
@@ -87,18 +87,42 @@ mod tests {
 
     #[test]
     fn check_project_with_workspace_context() {
-        let mut parser = ry_core::RParser::new().unwrap();
-        let file = parser.parse("test.R", "dplyr_function(1)\n").unwrap();
-        let mut workspace = ry_workspace::WorkspaceContext::default();
-        workspace.attached_packages.insert("dplyr".to_string());
-        let input = CheckInput {
-            files: vec![("test.R".to_string(), 0, Arc::new(file))],
-            user_stubs: Arc::new(BTreeMap::new()),
-            workspace: Some(workspace),
+        // `is_null(x)` narrows `x` away from NULL only when its defining
+        // package is attached, so the trailing `x()` is an RY070
+        // (calling a non-function) exactly when rlang is absent. Running
+        // the same source with and without the workspace context proves
+        // check_project actually feeds attached_packages into the checker
+        // instead of ignoring it.
+        let src = "x <- NULL\nif (is_null(x)) stop(\"missing\")\nx()\n";
+        let run = |workspace: Option<ry_workspace::WorkspaceContext>| -> usize {
+            let mut parser = ry_core::RParser::new().unwrap();
+            let file = parser.parse("test.R", src).unwrap();
+            let output = check_project(CheckInput {
+                files: vec![("test.R".to_string(), 0, Arc::new(file))],
+                user_stubs: Arc::new(BTreeMap::new()),
+                workspace,
+            });
+            output
+                .diagnostics
+                .iter()
+                .flat_map(|(_, diags)| diags.iter())
+                .filter(|d| d.code == "RY070")
+                .count()
         };
-        let output = check_project(input);
-        // Should complete without error even with workspace context.
-        assert!(!output.diagnostics.is_empty());
+
+        let without = run(None);
+        assert!(
+            without > 0,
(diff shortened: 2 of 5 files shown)


sims1253:
@coderabbitai review

PR #101, opened 2026-08-17
chore(analysis): delete unused host, snapshot, symbol, and catalog surface

sims1253:
## Summary

`ry-analysis` grew ~2,100 lines of infrastructure (the plan-38 analysis host and snapshot/symbol query engine, the plan-39 semantic catalog) that no consumer ever picked up. This deletes all of it. The crate is now `check.rs` plus a minimal `lib.rs`; its only consumer is (and was) `ry check` via `CheckInput`/`check_project`.

Closes #85 - the fidelity losses it tracks live in a conversion with no consumers.
Closes #40 - same for the schema fields feeding the dead catalog.

## Deleted

- `crates/ry-analysis/src/`: `catalog.rs` (285), `catalog_adapter.rs` (188), `effect.rs` (125), `interactive.rs` (67), `layering.rs` (87), `rules.rs` (85), `snapshot.rs` (400), `symbols.rs` (445), `tests/property.rs` (143), and the `AnalysisHost`/`Change` machinery, re-exports, and unit tests from `lib.rs` (262 lines).
- The `ry-analysis` dependency edge in `crates/ry-lsp/Cargo.toml` - ry-lsp's source contained zero references to the crate (it drives `ry_checker::Project` directly).
- Now-unused deps from `crates/ry-analysis/Cargo.toml`: `ry-config`, `tracing`, `tempfile`, `proptest`.
- `docs/architecture/analysis-query-engine.md`, `docs/architecture/p39-w1-catalog-design.md`, `docs/architecture/p39-acceptance.md`.

Net: 19 files, +38 / -2,349.

## Consumer verification

- ry-cli uses exactly two items from the crate: `CheckInput` (`main.rs:923`) and `check_project` (`main.rs:937`). Both kept.
- `rg ry_analysis crates/ry-lsp/` returns zero hits (src and tests).
- `check.rs` imports nothing else from the crate, so the deletion cannot be load-bearing.
- Full-tree grep for every deleted symbol (`AnalysisHost`, `SemanticCatalog`, `AnalysisSnapshot`, `SymbolIndex`, `layer_catalogs`, ...) hits no code. In docs, hits are the three deleted records plus historical mentions in the surviving plan-38/39 records, which now carry deletion notes.

Behavior-neutral by construction: the deleted code had no callers.

## Edits beyond deletion (review focus)

- `ry-analysis` Cargo.toml `description` named only deleted machinery; rewritten.
[...]

PR #103, opened 2026-08-17
feat(cli): add ry dump-types subcommand

sims1253:
## Summary

Adds a non-interactive CLI that dumps inferred type information per lexical scope, so downstream tooling can query "which bindings are in scope at line N, with what inferred types":

```
ry dump-types <FILE>... [--project-root <DIR>] [--format json] [--position LINE:COL]...
```

One or more R files, or a directory (expanded via `ry check`'s own discovery rules, so a dump and a check always agree on the file set). Output is JSON on stdout: files -> scopes (kind/name/start/end) -> bindings (name/kind/type/start). Deterministic order: scopes by start position, bindings by name. Types are `RType`'s Display strings - the same strings LSP hover returns - with the fully-uninformed type printed as `unknown`; one untyped binding never fails the run. One analysis pass that reuses `ry check`'s exact pipeline (config discovery, per-DESCRIPTION-package grouping, workspace resolution, shared fixpoint), so emitted types are what a check run infers.

## Implementation

Scopes are transient walk state in the checker - not retained after a check, with no positions or names - so this adds an opt-in capture path rather than changing retention:

- `ry-checker`: public `ScopeRecord`/`ScopeRecordKind` types; `capture_scopes` flag on `Checker` (off by default) with `enable_scope_capture`/`take_scope_records`. Function scopes snapshot at the end of `walk_stmt`'s function-body arms, gated on `capture && !discarding` so fixpoint/signature re-walks never double-capture; the top-level scope is recorded at the end of `emit_diagnostics`.
- `ry-checker::Project`: records ride the existing per-file tuple through the pass-3 emitter (no Mutex), with `enable_scope_capture`/`take_scope_records` wrappers.
- `ry-analysis`: `check_project_with_scope_capture`, same wiring as `check_project` via a shared private helper.
- `ry-cli`: the subcommand - parallel parse (existing thread-local parser pattern), grouping, JSON assembly, binding-kind classification, AST scan for definition sites, position filtering.

Capture is off by default, so `ry check` and the LSP paths are untouched (existing suites unchanged).

## Binding kinds

`param` = formal of this scope still marked as parameter (a reassigned formal degrades to `local`, matching R's rebinding). `local` = first assigned in this scope's body (blocks included; R has no block scope). `closed-over` = function-scope binding inherited via the enclosing-scope clone. `imported` = top-level binding never assigned in the file (ambient, e.g. Shiny server fragments). Documented in the README section.

## Documented limitations (README + code comments)

- Anonymous function literals in argument position are not recorded as scopes - the checker only ever walks them in discarding mode (checker design, not a dump limitation).
- Snapshots are end-of-body state; a nested scope reflects the enclosing scope at the inner function's definition point (ry's documented closure approximation). `--position` compensates by dropping locals first assigned after the position.
- `--format` currently accepts only `json`.

## Tests

- `crates/ry-cli/tests/dump_types_e2e.rs` (4 tests): JSON shape, a param binding with a concrete type, an `unknown` case, `--position` filtering, exit-code semantics (0 with diagnostics present; 1 for usage-adjacent IO/config failures; 2 for clap usage errors).
- Unit tests in `project.rs`, `check.rs`, and the CLI (classification, capture gating).
- The integration suite caught a real bug during development: `Scope::clone` copies the parameter-marker set, so a captured outer formal would misclassify as `param` of the inner scope; classification now requires declaration in that record's own formals.

## Performance

Release build on a 300-file package tiled from vendored purrr sources: `ry check` 0.50-0.59s vs `ry dump-types` 0.54-0.57s - parity within noise; one pass, no per-position re-analysis.

## Smoke tests (trimmed)

- `glue/R/safe.R`, full dump, exit 0: top scope locals `glue_safe`/`glue_data_safe`/`get_transformer` as `function<len=1>`; inside `glue_safe`: params `...`/`.envir` (`unknown`), `glue_safe` closed-over.
- `purrr/R/adverb-partial.R --position 98:10`, exit 0: one `function`/`partial` scope `[82,12]`-`[140,2]`; params `.f`/`...` `unknown`; locals `env` -> `opaque<len=?>`, `heterogeneous_envs` -> `logical<len=1>`; `partial` closed-over; `fn_sym` (first assigned line 105, after the position) correctly absent.

## Merge order

Touches `crates/ry-analysis/src/check.rs` and `lib.rs`, which #101 also edits (doc corrections + lib.rs reduction). Rebase this on main after #101 lands; the conflict is textual and small.

## Gates

- `cargo test --workspace` - 45 suites ok, 0 failures (includes the R oracle gate, R 4.6.1)
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo fmt --all -- --check` - clean

## Summary by CodeRabbit

* **New Features**
  * Added the `ry dump-types` command to output lexical scopes and inferred bindings as JSON.
  * Supports project-root selection, directory discovery, source-position filtering, and nested scopes.
  * Reports binding kinds, definition locations, parameters, closures, imported names, and inferred types.
  * Provides distinct exit handling for usage, I/O, configuration, and analysis issues.
  * Supports opt-in lexical-scope capture during project analysis without changing existing commands.

* **Documentation**
  * Added README and changelog documentation covering output format, filtering, project discovery, and command behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to ry are documented in this file.
 
 ## [Unreleased]
 
+### Added — `ry dump-types` CLI
+
+- **Non-interactive type dump**: `ry dump-types <FILE>...` runs one
+  analysis pass (identical pipeline and environment to `ry check`) and
+  prints every lexical scope of the requested files as JSON on stdout:
+  scope kind/name/extent plus each binding's name, kind
+  (`param`/`local`/`closed-over`/`imported`), type string (the same
+  `Display` rendering the LSP hover returns; `unknown` when inference
+  has nothing), and definition site. `--position LINE:COL` (repeatable)
+  restricts output to the innermost scope containing each position and
+  drops locals assigned after it. `--project-root <DIR>` overrides the
+  analysis root for non-package files; the default mirrors `ry check`'s
+  per-package (DESCRIPTION) grouping. Exit code stays 0 even when the
+  analyzed code has diagnostics; non-zero only for usage, IO, or
+  internal failure.
+- **Opt-in scope capture in the checker**: `Checker` and `Project` can
+  snapshot every completed lexical scope during the pass-3 diagnostic
+  walk (`ry_checker::ScopeRecord`); recording is suppressed in
+  discarding mode so fixpoint and signature-building re-walks never
+  double-capture. `ry_analysis::check_project_with_scope_capture`
+  exposes it through the shared check pipeline. Existing commands and
+  checker behavior are unchanged.
+
 ### Plan 37: Release truth and editor hardening
 
 #### P37-W1: Parser UTF-8 boundary panic fix
--- a/README.md
+++ b/README.md
@@ -245,17 +245,21 @@ ry dump-types R/analysis.R --position 4:5
 ```
 
 A directory argument expands to every discoverable R file under it,
-using `ry check`’s discovery rules. `--project-root <DIR>` overrides the
-analysis root for non-package files; by default each file is analyzed in
-the context of its nearest enclosing package (the ancestor directory
-with a `DESCRIPTION`), else the working directory — mirroring `ry check`’s
-per-package grouping. The exit code is 0 even when the analyzed code has
+using `ry check`’s discovery rules, including the discovered `ry.toml`’s
+`exclude` patterns. `--project-root <DIR>` overrides the analysis root
+for non-package files; by default each file is analyzed in the context
+of its nearest enclosing package (the ancestor directory with a
+`DESCRIPTION`), else the directory owning the discovered `ry.toml`, else
+the working directory — mirroring `ry check`’s per-package grouping. The
+exit code is 0 even when the analyzed code has
 diagnostics; it is non-zero only for usage, IO, or internal failure.
 Scopes reflect the checker’s snapshot semantics: each table is the
 scope’s state at the end of its body, and a nested function captures the
 enclosing scope as of its definition point (ry’s documented closure
 approximation). Anonymous function literals used as call arguments are
-inferred in discarding mode and are therefore not recorded as scopes.
+inferred in discarding mode and are therefore not recorded as scopes —
+but named functions defined *inside* such a callback do complete and are
+recorded, so a dump can contain a scope whose enclosing scope is absent.
 
 ## Configuration (`ry.toml`)
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,85 @@ functions do not produce false unbound-variable reports. When the
 masked data’s schema is unknown, ry stays silent rather than guessing
 at column candidates.
 
+## Dumping inferred types
+
+`ry dump-types` is the non-interactive counterpart of the LSP hover: it
+runs one analysis pass (the same pass `ry check` runs, over the same
+package-aware environment) and prints every lexical scope of the
+requested files as JSON on stdout. Downstream tooling can query which
+bindings a scope holds and with what inferred types, without re-running
+the checker per position.
+
+``` bash
+ry dump-types R/analysis.R
+#> {
+#>   "files": [
+#>     {
+#>       "path": "R/analysis.R",
+#>       "scopes": [
+#>         {
+#>           "kind": "function",
+#>           "name": "moving_average",
+#>           "start": [3, 19],
+#>           "end": [8, 2],
+#>           "bindings": [
+#>             { "name": "n", "kind": "param", "type": "integer<len=1>", "start": [3, 31] },
+#>             { "name": "threshold", "kind": "closed-over", "type": "integer<len=1>", "start": [1, 1] },
+#>             { "name": "window", "kind": "local", "type": "function<len=1>", "start": [4, 3] },
+#>             { "name": "x", "kind": "param", "type": "unknown", "start": [3, 28] }
+#>           ]
+#>         }
+#>       ]
+#>     }
+#>   ]
+#> }
+```
+
+Positions are 1-based `[row, column]` pairs; columns count characters,
+not bytes. Scopes are ordered by start position, bindings by name.
+`type` is the same string the LSP hover shows; `unknown` marks bindings
+ry could not infer, and never fails the run.
+
+Binding kinds:
+
+- `param` — a formal of this scope that the body never reassigns. A
+  reassigned formal degrades to `local` at its reassignment site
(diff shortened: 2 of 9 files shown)


sims1253:
Review feedback triage (commit 3e834c0). Every point was verified against the code and the built binary before acting.

**Contract divergence 1: discarded ry.toml path / exclude patterns not applied (main.rs:1531, main.rs:1545) — fixed.**
Verified real: with `ry.toml` containing `exclude = ["**/vendor/**"]`, `ry check .` analyzed 1 file while `ry dump-types .` dumped 2 (vendor file included). Also confirmed `discover_r_files` documents `exclude_root: None` as "exclude patterns are not applied", and `run_check_once` falls back to the config root for non-package groups where the dump fell back to cwd. Fix: `run_dump_types` now keeps the config directory as `config_root`, passes it as `exclude_root`, and inserts it between `--project-root` and cwd in the resolution-root chain (`--project-root` still overrides). Re-verified: check and dump now report the same single file. Test: `directory_dump_applies_config_exclude_patterns_like_check` dumps only the non-excluded file and asserts check's "checked 1 file(s)" summary over the same directory.

**Contract divergence 2: dedup on span.start alone can delete the top scope (main.rs:1310) — fixed.**
Verified real: for `function(x) x` followed by `a <- 1L`, the dump contained only the anonymous function scope; the `top` scope and its `a` binding were gone (the function record and `whole_file_span` both start at byte 0, and the stable sort kept the function record first). Fix: the dedup key is now `kind` plus start offset, preserving the defense against re-walk duplicates. Test: `leading_function_literal_keeps_top_scope_and_its_bindings` asserts both scopes, the `a` binding, and that `--position 2:1` selects the top scope.

**Nitpick: README missing the nested-named-function-in-callback caveat — fixed.**
Verified real: `lapply(1:3, function(item) { helper <- function(v) v + 1L; helper(item) })` dumps the `top` and `helper` scopes but not the callback. The README limitations paragraph now says named functions defined inside such callbacks are recorded while the callback itself is not. Test: `named_function_inside_anonymous_callback_is_recorded` pins the documented shape (`["top", "helper"]`).

**Nitpick: degraded_scopes dropped silently — fixed.**
Verified real: the dump never read `package_scope.degraded_scopes` where check prints a summary. The dump now prints one stderr note per degraded scope with check's wording; stdout JSON is untouched. Test: `degraded_scopes_are_noted_on_stderr_only` (LazyData package, oversized data/big.rda, `max-serialized-bytes = 16`).

**Nitpick: innermost-scope scan ran twice — fixed.**
Verified by reading: the two loops were identical. `assemble_file_dump` now runs one pass that records both the selected scopes and their selecting offsets.

Gates from the worktree root: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --all -- --check` all pass.

sims1253:
Review response to the 2026-08-18 CodeRabbit round:

**1. Walker recursion (functional correctness, major) — fixed in 4a015d2, with e2e tests.** Verified against the built binary before changing anything: `print(side <- 5L)` dumped `side` as `imported` with `start: null`; `outer <- inner <- 1L` missed `inner`; a named function inside an anonymous callback got no `function_locals` (its own local dumped as `closed-over`, null site); `if (flag <- TRUE)` missed `flag`; `q %<>% abs()` dumped `q` as `imported`. Both `*_in_expr` walkers now recurse through Call (func and args), BinOp (lhs and rhs), UnaryOp, Index (base and args), Block, and If; the statement walkers additionally cover If conditions, For iterables, While conditions, and Return values — the same recursion set as ry-checker's `visit_expr`/`visit_stmts`. Expression-position assignment recording covers Assign/SuperAssign and PipeAssign (`%<>%` rebinds its LHS in the checker's BinOp arm; verified no RY010 on a later read). Assign targets are deliberately not walked: the checker leaves `m[i <- 1L] <- v` unbound (RY010 on later use, verified), and the dump reports what `ry check` infers. Tests: `expression_position_assignments_are_locals_with_sites` (call argument, chained assignment, if condition, `%<>%`, and --position filtering on the new sites) and `named_function_inside_loop_body_callback_gets_locals` (named function inside an anonymous callback inside a for body: own locals plus closed-over sites resolved through the unrecorded callback). The "heavy lift" label was overstated — the fix stays inside the two dump-layer walkers.

**2. Enclosing-chain precompute (performance) — fixed in 38e4e16.** `enclosing_binding_span` allocated and sorted the candidate scope list once per closed-over binding; `enclosing_scope_chains` now computes every scope's nearest-first chain once per file and lookups scan the cached chain. Same filter and ordering, so resolution output is unchanged (asserted via the closed-over sites in the loop-body test).

**3. --project-root coverage (maintainability) — fixed in 702c046, with one deviation from the suggested assertion.** The suggested mechanism (a binding defined in a sibling file resolving through the flag) does not exist: cross-file name pooling for non-package scripts is resolution-root-independent, and `load()` inventory enumeration only runs for files inside a package root (it lives in `resolve_workspace_context`'s package branch) — verified with an instrumented run showing `load_bindings` empty for non-package files regardless of root. For non-package groups the root currently only gates workspace-root validation. The new test covers the flag's actual contract: two sibling non-package script directories under a shared parent whose `ry.toml` is discovered from the first input — the config-root fallback alone succeeds, `--project-root` at the same parent behaves identically, and pointing the flag at a non-directory fails with "workspace root is not a directory", proving the flag takes precedence over the config root.

Gates: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all clean.

sims1253:
Review feedback triage (commit 91ae486), both points verified against the built binary before acting.

**Assert the inclusive --position boundary — fixed.** Verified: the test described `outer` starting exactly at the queried position (visibility rule is `span.start <= offset` in `assemble_file_dump`) but never asserted it, so a regression to a strict `<` would pass. Assertion added in 91ae486.

**Assert cross-file visibility via the shared resolution root — declined, the asserted behavior does not exist.** Tried the suggested assertion first: it fails. `b.R`'s top scope contains exactly one binding — `{"kind":"local","name":"b_val","start":[1,1],"type":"opaque<len=?>"}" — and no `a_val`. This matches the instrumented finding from the previous round: for non-package groups, cross-file binding pooling never runs (`load_bindings` stays empty regardless of root), so `--project-root` controls root validation and grouping, not binding visibility. Pinning `a_val` in `b.R`'s dump would encode a contract the checker does not implement. If cross-file visibility for pooled non-package scripts is wanted, that is a checker feature request, not a test gap.

Issue #90, opened 2026-08-14
PR #79 review leftovers: triaged as real, never implemented


sims1253:
Findings from the PR #79 review rounds that were triaged as real but never
implemented. I queued them for a worker that was never dispatched, so they would
be lost on merge. Verified still present at `52d840c`.

Marked **[verified]** where I confirmed against the code just now.

## Correctness

- **[verified]** `crates/ry-analysis/src/lib.rs:130` — `Change::SetOpenFile`
  does `self.open_files.insert(path, (version, content))` with no version
  comparison. An out-of-order update (version 1 arriving after version 2)
  silently replaces newer source with older. Retain only a strictly newer
  version for an already-open path, and add an out-of-order test.

- **[verified]** `crates/ry-checker/src/project.rs:247` —
  `force_full_recollection` clears `collected_files` before any `update_file`
  can move the old function names into `invalidated_fns`, so
  `compute_fixpoint_scope` sees only the new definitions. Transitive callers of
  a removed or renamed function can keep stale return types and
  caller-visible signatures. Preserve the names before clearing:

  ```rust
  self.invalidated_fns.extend(
      self.collected_files.values()
          .flat_map(|previous| previous.fn_table.fns.keys().cloned()),
  );
  self.collected_files.clear();
  ```

  Note this interacts with #81 — `force_full_recollection` may be removed
  entirely once that is resolved, which would make this moot. Check #81 first.

- **[verified]** `crates/ry-analysis/src/snapshot.rs:75,79` — `files` and
  `diagnostics` are `HashMap`, so traversal order is unstable and leaks into
  snapshot and symbol-query results. Use `BTreeMap`, or sort paths before
  building indices. Same for `crates/ry-analysis/src/symbols.rs` (~124-139):
  sort references by file path and byte range before returning.

## Test quality

- **[verified]** `editors/zed/src/lib.rs:455` — `cached_path_precedence` only
  calls `GithubReleaseDetails::new`. It never sets `cached_binary_path` and
  never calls `language_server_binary`, so it passes even if cache precedence
  regresses. Either rename it to describe path construction, or add a test that
  actually exercises the cache branch. Fits the pattern in #82.

- `crates/ry-lsp/tests/p36_contract.rs:1363` — the drain loop stops at the
  first timeout, so a single 500 ms silence window can end it before background
  indexing plus the 180 ms debounce have published. Both vectors stay empty and
  the assertions fail with no production defect. Keep draining until both files
[...]

PR #105, opened 2026-08-18
fix: address surviving #90 review leftovers

sims1253:
## Summary

Addresses the surviving findings from #90 (PR #79 review leftovers triaged as real but never implemented). The other two #90 items (SetOpenFile version guard, snapshot/symbols ordering) live in crates/ry-analysis code that #101 deletes, so they ride with that PR and are out of scope here.

Part of #90.

## 1. Zed test theater - fixed (test(zed): pin binary-resolution cache precedence)

`cached_path_precedence` only checked that a constructed path ends in `ry` - path construction is already exhaustively covered elsewhere. Binary resolution now decides through an extracted pure `BinarySource` precedence function plus a `cached_binary_on_disk` check, and new tests pin the full order: settings > PATH lookup > cache > download, with the cache entry backed by a real temp file. Mutation check: deleting the cache branch fails `binary_source_precedence`; the old test passed under the same mutation (the #82 failure mode).

`Worktree`/`LspSettings::for_worktree` are host-IPC backed (WIT re-exports), so end-to-end `language_server_binary` coverage is not feasible in unit tests. One disclosed nuance: `worktree.which("ry")` is now evaluated even when a settings path is configured - a pure, read-only host query whose result is discarded.

## 2. p36_contract drain loop - fixed, harness only (test(lsp): drain p36-w6 until sentinel files publish)

The issue text maps to the w6 loop (`first_diags`/`last_diags`, file_00/file_31), not the w7 loop the drifted line number pointed at. The loop now drains until both sentinel files publish, bounded by a 30 s total deadline via `timeout_at`, replacing the first-500-ms-silence termination. The quiescence tail is gone too. No production code touched. Verified: 5 consecutive runs all 10/10, including 2 under 8-way CPU load on an already-loaded machine (load avg ~20).

## 3. check-ledger.py validation gap - fixed (fix(ecosystem): reject offsetting ledger summary corruption)

Demonstrated the gap first: the old script accepted a `classification` map doctored with `{ghost_label: -1, phantom_label: +1}` - totals and observed counts preserved. `compare_summary` now walks the union of summary and observed keys. Key discovery: strict map equality would be wrong - the committed posit-0.9.0.json legitimately carries `uncertain: 0` (fixed taxonomy), so zero-valued entries for unseen categories stay valid; every other mismatch (extra non-zero, negative, missing, drift) fails. Validated against the committed posit-0.9.0.json and posit-0.8.0.json, which are the only CI callers (two ecosystem.yml invocations).

## 4. Dropped degraded_scopes in ry-cli - declined as already handled

Two-part evidence:
- The reporting exists: `package_scope.degraded_scopes` is collected into `CheckResult.degraded` and printed by `print_degraded` in both the human summary and `--statistics` (landed in #39, before #90 was filed). Verified end-to-end with a package fixture carrying a >2 MiB-decoded `.rda` (default cap is 2 MiB): `ry check` prints the note, `--output-format json` stays clean by design.
- The literal `degraded_scopes: Vec::new()` into `WorkspaceContext` is inert - `check_project` never reads that field, so forwarding it would be a dead assignment with no observable effect.

No CHANGELOG entry for this item (no behavior change).

## New findings posted back to #90

Three items surfaced during verification that belong on the issue rather than in this PR - see the #90 comment: the w7 drain loop's false-PASS silence-window weakness (needs a design decision), the dead `WorkspaceContext.native_registrations` plumbing, and a stale file-header claim in p36_contract.rs.

## Gates

- `cargo test --workspace` - 1004 passed, 0 failed, 44 binaries (includes the R oracle gate)
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo fmt --all -- --check` - clean
- `cargo build` + `cargo test --manifest-path editors/zed/Cargo.toml` - 8 passed, 0 failed
- `cargo test -p ry-lsp --test p36_contract` - 5 runs, all 10/10

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved language-server binary selection by prioritizing configured paths, system-installed binaries, valid cached files, and downloads.
  - Prevented stale or invalid cached binaries from being selected.
  - Improved diagnostics collection for large, multi-file language-server projects.
  - Strengthened ledger validation to detect incorrect counts for newly observed categories.

- **Documentation**
  - Updated the unreleased changelog with these fixes and validation improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,26 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- The Zed extension's cache-precedence test asserted only that a
+  constructed release path ends in "ry" (#90, the #82 test-theater
+  family), so it passed with the cache branch deleted outright. Binary
+  resolution now runs through an extracted `BinarySource` precedence
+  function whose tests pin settings > PATH lookup > previous download >
+  download and require the cached file to still exist on disk
+  (editors/zed/src/lib.rs).
+- The P36-W6 many-files test's drain loop stopped at the first silent
+  500 ms window (#90), so a loaded machine could end collection before
+  the background index plus the publish debounce had emitted anything,
+  failing the sentinel assertions with no production defect. The loop
+  now drains until both sentinel files have published, bounded by a
+  total deadline instead of a per-message gap
+  (ry-lsp/tests/p36_contract.rs).
+- `ecosystem/check-ledger.py` compared each summary block per observed
+  key only (#90), so a negative entry for an unseen label offset by a
+  positive one preserved both the total and every observed count. Each
+  summary is now compared over the union of summary and observed keys;
+  zero-valued entries for categories no finding carries remain valid
+  (the classification taxonomy is fixed), and any other mismatch fails.
 - Tests that could not fail were fixed or deleted (#82): the ry-analysis
   workspace-context test now asserts the rlang/RY070 flip in both
   directions (check.rs), the same-name symbol test asserts reference
--- a/crates/ry-lsp/tests/p36_contract.rs
+++ b/crates/ry-lsp/tests/p36_contract.rs
@@ -1267,20 +1267,30 @@ fn p36_w6_many_files_flat_filter_construction() {
             "textDocument": {"uri": trigger_uri, "languageId": "r", "version": 1, "text": trigger_text}
         })).await.unwrap();
 
-        // Drain ALL diagnostic publications with a timeout-based approach
-        // (like p36_w7). The background index + debounced publish will send
-        // diagnostics for trigger.R and all indexed files. We collect
-        // diagnostics for file_00 and file_31 from the drain.
+        // Collect diagnostics for the first and last indexed files from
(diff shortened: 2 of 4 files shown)


sims1253:
Addressed the `cached_binary_on_disk` directory-path finding in 8e45731.

Verified the claim: the guard `fs::metadata(path).is_ok_and(|stat| stat.is_file())` does reject a directory, but `cached_binary_on_disk_requires_a_file` only pinned existing-file and missing-path cases, so a regression to a metadata-existence check (`fs::metadata(path).is_ok()`) passed the suite.

Fix: extended the test to assert a directory at the cache path is rejected, per the review suggestion. Mutation check confirms the regression is now caught: with the guard reduced to an existence check, the test fails at the new directory assertion.

Gates: cargo test --manifest-path editors/zed/Cargo.toml (8 passed), cargo build --manifest-path editors/zed/Cargo.toml, cargo fmt --all -- --check, cargo clippy --workspace --all-targets -- -D warnings, cargo test --workspace all green.

PR #115, opened 2026-08-24
ci(ecosystem): validate --check against a clean checkout (#50)

### LINKED ISSUE #50: Ecosystem --check must validate against a clean checkout
## Problem

Changes to `ecosystem/run.sh`'s drift-check logic are validated against the developer's working tree, which may contain untracked `.full.txt` files from prior runs. This allowed a broken change (checking gitignored `.full.txt` baselines) to pass locally and fail only in CI.

## Fix

Add a CI step or local pre-push hook that validates `run.sh --check` against a clean checkout state:

1. Remove all gitignored report files before running `--check`: `rm -f ecosystem/reports/*.full.txt`
2. Assert `--check` exits 0 after the cleanup (proving the committed `.txt` files are the sole baseline)
3. Then run `test-drift-detection.sh` (which corrupts committed files)

This makes it impossible to accidentally add a comparison against a file that doesn't exist in the repo.

sims1253:
Closes #50.

## Problem

Changes to `ecosystem/run.sh`'s drift-check logic were validated against the developer's working tree, which can carry gitignored `ecosystem/reports/*.full.txt` files from earlier update runs. A broken change that compared against those gitignored baselines passed locally and failed only in CI.

## Change

Two edits in `.github/workflows/ecosystem.yml`:

1. **PR-time clean-state validation** — one step inserted in the existing `ecosystem` job, between the release build and the snapshot check:
   - `rm -f ecosystem/reports/*.full.txt`
   - the existing `run.sh --check` now doubles as the exit-0 assertion on a clean state (proving the committed `.txt` snapshots are the sole baseline)
   - the existing drift-detection integration test follows immediately, as before
2. **Drop the `clean-checkout` job** — it was schedule/dispatch-only, so its fresh-clone build was never consumed; the PR-time step above delivers the actual protection it aimed at, on every PR instead of on a cron nobody watched.

For reference: `--check` generates reports into a throwaway temp dir and `cmp -s`s them against the committed per-package `.txt`/`.root.txt` snapshots and summaries; the message-bearing `.full.txt` variants are never compared. The committed logic is already correct — this makes it impossible for future drift-check changes to pass by leaning on files absent from a fresh checkout.

## Verification

Reproduced (1) locally against the real drift logic (`run.sh --local` with a prebuilt binary at the same commit):

- stray gitignored `.full.txt` files present → confirmed gitignored; current logic ignores them
- after the cleanup step → `--check` exits 0 ("committed reports are current")
- corrupted a committed baseline → `--check` exits 1 with report drift; restored
- full `test-drift-detection.sh` → PASS

And (2) for the job removal: YAML parses (`pyyaml`), remaining jobs (`ecosystem`, `posit`, `cargo-edges`) intact. The workflow itself only fully executes on GitHub.

Not run locally: the full non-local `--check` over all ~90 manifest packages (needs network clones).

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -103,34 +103,6 @@ jobs:
         run: ecosystem/test-posit-drift-detection.sh
       - name: Posit label falsification integration test
         run: ecosystem/test-posit-label-falsification.sh
-  # P37-W7b: Clean-checkout validation gate (#50)
-  # Validates that the repo is releasable from a fresh clone with no
-  # gitignored artifacts.
-  clean-checkout:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          persist-credentials: false
-      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
-      - name: Remove all gitignored files
-        run: git clean -fdX
-      - name: Build release binary from source
-        run: cargo build --release --locked -p ry-cli --bin ry
-      - name: Validate ledger consistency
-        run: python3 ecosystem/check-ledger.py docs/corpus/posit-0.9.0.json
-      - name: Verify no ignored artifact is required
-        run: |
-          set -euo pipefail
-          # Verify that after clean -fdX, the repo still builds and the
-          # ledger is valid. If this fails, an ignored file is required
-          # for a gate, which means a release-time clean checkout breaks.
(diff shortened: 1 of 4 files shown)


sims1253:
Addressed the review in 2442afd:

- **The inert CI step / #50's real failure mode**: `test-drift-detection.sh` now plants a poisoned `ecosystem/reports/glue.full.txt` sentinel (trap-cleaned) before its first `run.sh --local --check` assertion. Verified both directions locally: current logic ignores the sentinel and the script passes; a run.sh patched to prefer `.full.txt` baselines aborts at that first check with exit 1 — i.e. the #50 recurrence path now fails on a developer machine too. With the behavioral pin landed, closing #50 on merge is back to being justified; the workflow `rm` step stays as an explicit statement of the invariant.
- **Orphaned doc references**: the release-runbook item now points at every PR's fresh-checkout `ecosystem` build instead of the deleted scheduled job. The `project-remediation-plan.md` citation is left to #117, which deletes that file wholesale.

Issue #130, opened 2026-08-31
apply_narrowing clones the whole Scope twice per if — snapshot with an undo log instead


sims1253:
## Why

rustc's inference context journals every unification mutation into a `Vec<UndoLog>`; a snapshot is just a length into that vector, and rollback replays **only the changes**. That is what makes speculative, branch-heavy inference affordable.

ry clones instead: `apply_narrowing` (`crates/ry-checker/src/infer/misc.rs:495`) does `(base.clone(), base.clone())` per `if`, and closures enter via `scope.clone()` (`crates/ry-checker/src/infer/mod.rs:313`). A `Scope` is a `HashMap<String, RType>` **plus five side-sets** (`narrowed_bindings`, `parameter_bindings`, `list_origin_bindings`, `function_aliases`, `unreachable` — see `lib.rs:265-380`), so each clone is O(scope size), and nested if-chains compound it: work grows with depth × scope size, which is the quadratic smell this issue removes.

## Proposal

- Journal scope mutations as `(key, old_value)` entries — one shared journal across the maps, or an `im::HashMap` if the ergonomics win.
- Snapshot = journal length (O(1)); branch analysis = mark, mutate in place, then **commit or roll back to the mark**. `then`/`else` share one base with two journals instead of two deep clones.
- `merge_branch_bindings` keeps its join semantics unchanged — this is a storage change, not a semantics change.

## The cautionary tale that comes with it

rustc's worst next-solver regression (unic-ucd-name, **214x slower**) was precisely snapshot rollback iterating *all* union-find variables even when nothing had been touched; making rollback proportional to actual changes took it to 1.07x (rust-lang/rust#160982). The lesson for us: fix O(scope) *snapshot* without introducing O(scope) *rollback* — the journal must be the source of truth for what changed, not a diff reconstructed from the full state.

## Acceptance

- [ ] Snapshot cost is O(1); rollback proportional to mutations since the mark
- [ ] Narrowing/merge fixture behavior identical: all 384 corpus fixtures + oracle harness green
- [ ] Allocation reduction on a nesting-heavy fixture visible in the criterion bench
- [ ] No behavior visible outside `infer/` — the `Scope` API change should not ripple into rules or LSP hint consumers

## Sources

- [rustc_infer undo_log docs](https://doc.rust-lang.org/stable/nightly-rustc/rustc_infer/infer/snapshot/undo_log/index.html)
[...]

PR #137, opened 2026-08-31
refactor(checker): mechanical scope-clone and duplication cleanup in infer

sims1253:
## Summary

Mechanical prefix of #130. It removes gratuitous scope clones, dead
parameters, and duplicated helper logic inside `crates/ry-checker/src/infer`.
Semantics do not change. The undo-log redesign of scope storage stays
tracked in #130.

## Changes

- `apply_narrowing` drops its unused `_has_else` parameter and returns
  early for `Narrowing::None`, so the common no-narrowing `if` skips the
  match-arm work while returning exactly what it returned before.
- `merge_branch_bindings` takes the branch scopes by reference. The call
  site in `walk_stmt` no longer clones both scopes just to keep reading
  them.
- `infer_if_expr` no longer clones the branch scopes it consumes last.
- A single `is_comparison` predicate replaces the `Lt | Le | Gt | Ge | Eq | Ne`
  list written out at six sites (binop, call, recall, index).
- A single `concrete_json_mode` mapping replaces the JsonMode-to-Mode
  table written out three times (`json_rtype_scalar`, union member
  filtering, `apply_sig`). Each caller keeps its own fallback for
  non-concrete specs.
- `extract_builtin_type_narrowing` shares the `inherits` target lookup and
  the first-argument identifier extraction.
- `apply_narrowing` shares the narrow-away-from-null sequence that was
  repeated across four arms.
- `try_s3_binop_dispatch` and `try_s3_unary_dispatch` share the class
  vector walk.
- `equality_list_leaf_type` loses its dead `op` parameter.
- Unused `span` and `name` parameters are underscore-named at the
  signature, matching the sibling `_span` convention, instead of silenced
  with `let _ = x;` in the body.
- The `t` callee check uses plain string equality like its neighbors, and
  the data-mask lookup uses `.filter(...)` instead of a re-borrow plus
  `expect`.
- A `use` floating mid-file in misc.rs moves to the top, and a stale doc
  sentence on `op_symbol` (the match is exhaustive) is gone.

## Verification

- `cargo fmt -p ry-checker`
- `cargo test -p ry-checker` — full crate suite, all green: 460 unit
  tests plus every integration suite (corpus, metamorphic, oracle,
  project, recall rules, real world, snapshots), 0 failures.


sims1253:
Dropped the dead `symbol == "?"` guard in `try_s3_binop_dispatch` (4b6f279): `op_symbol` is exhaustive over `BinOpKind` and never returns `"?"`, so the clause could not fire. Full `ry-checker` suite passes locally (595 tests, 0 failures); clippy and fmt clean.

PR #144, opened 2026-08-31
refactor(lsp): one filter implementation, shared republish, leaner publishes

sims1253:
## Summary

Removes structural duplication and hot-path clones the base PR left alone. No behavior change. Stacked on #140; merge after it.

## Changes

**One filter implementation.** Deleted the test-only `State::merge_filter`, a near-verbatim copy of `compute_folder_filter` minus excludes and the counter. `effective_filter` now calls `compute_folder_filter` and keeps the filter. Also dropped the `file_config` parameter: every call site passed the same config twice, so the folder config is now both the exclude source and the severity fallback by construction.

**One republish path.** Extracted `republish_all_open_documents` and use it in `did_change_configuration` and `did_change_watched_files`. `did_change_workspace_folders` keeps its filtered republish on purpose: documents under removed roots were just cleared, they fall back to root-level eligibility, and a blanket reschedule would re-publish check results for them. A comment at the site states this.

**Legacy state removed (item 4 decision: collapse).** Dropped the `workspace_folders: Vec<(PathBuf, Config)>` field. Every write derived it from `folder_contexts` inside the same lock section (initialize, folder mutation, watched-files reload), so the root sets were identical at all times. That made both readers unreachable: the `eligibility_for_path` fallback only ran when no context root matched the path — the same root set could not match there either — and the `spawn_background_index` branch only ran when `folder_contexts` was empty, which implied the derived vector was empty too. There is no startup window where the vector is populated and the contexts are not; the two are written together.

**Leaner publishes.** Four trims on the path that runs after every debounced edit:
- Snapshot only the eligible documents' versions, not the whole `versions` map.
- Clone each disk file once, straight into the check input. The old path cloned into a HashMap and then cloned again.
- Pass each folder partition to `check_with_workspace` by value. It already re-collects its own snapshot.
- Consume the check results so each file's diagnostic vec is taken, not cloned.

Generation guards and version stamping are untouched.

**`scope_for` (item 6 decision: simplified, single check).** Dropped the pre-check `parsed_version` capture; one post-check remains: the parse cache still holds this exact `Arc` and its tag equals the live document version. Interleavings considered:
- Edit lands between the parse and the post-check: `update_doc` removes the entry, `Arc::ptr_eq` fails, no cache write. Same outcome before and after.
- A concurrent re-parse replaces the entry: the new entry holds a fresh `Arc`, identity fails, no cache write. Same outcome.
- Entry removed before the post-check and an identical-tagged entry installed after: impossible. Entries are only written by `record_parse` with a freshly parsed `Arc`, and the cache-hit fast path returns the stored `Arc` without re-installing. A superseded `Arc` never returns. So the pre-check could only re-derive what the post-check already proves, and the stored set is identical.
- The scope returned to the caller can be stale in both versions identically; the guard only protects the cache write.

**Test sleeps.** Removed the two 1s settles in `configuration_refresh.rs` and the 300ms settle in `fallback_isolation.rs`. The background indexer publishes nothing on completion and open documents shadow disk files, so the tests' existing bounded waits are the only synchronization they need. The `fallback_isolation` hint wait gains a 2s timeout so a failure stays diagnosable.

No CHANGELOG entry: internal only.

## Verification

- `cargo fmt -p ry-lsp`
- `cargo clippy -p ry-lsp --all-targets -- -D warnings`
- `cargo test -p ry-lsp` full suite green: 50 unit, 36 integration (including the folder-mutation convergence tests that pin the removed-root clearing the filtered republish preserves)
- The two former sleep-test files ran 5/5 stable after the change


sims1253:
Reconciled the two comments in 875c5f4. `spawn_background_index` never publishes diagnostics itself; it only stores results in `state.disk_files`. The callers await it and then republish (`did_change_watched_files`, `did_change_configuration`, `did_change_workspace_folders`), and that republish is what makes cross-file calls into unopened files resolve. The function doc now says exactly that, and the `configuration_refresh.rs` comment mirrors the same clause instead of sounding like the opposite claim.

On the UNSTABLE merge state: no check failed. `Oracle (R)` was still running (pending), `Macroscope - Correctness Check` is SKIPPED by design, CodeRabbit reports pass (review rate limited), and the statusCheckRollup contains one null/blank entry — a GraphQL artifact, not a check. UNSTABLE followed from the pending Oracle job alone.

PR #145, opened 2026-08-31
refactor(checker): unify function-entry and condition-diagnostics duplication

sims1253:
## Summary

Collapses the remaining large duplications inside `crates/ry-checker/src/infer/`. Diagnostics and inferred types are unchanged. This PR is stacked on #137 and should merge after it.

## Changes

- **One function-entry path.** `enter_function_body` replaces the two ~55-line body walks in `walk_stmt`: the `Stmt::Assign { value: Expr::Function }` arm and the `Stmt::FunctionDef` arm. The only line that differed is the S3 dispatch context, installed for the assignment form alone, so the helper takes the binding name plus a `named_binding` flag.
- **One condition-diagnostics emitter.** `emit_condition_diagnostics` replaces the RY001/RY003/RY002 block repeated in the `Stmt::If` arm, the `Stmt::While` arm, and `infer_if_expr`. A `ConditionContext` picks the message noun and carries the RY002 case. On the While/RY002 divergence: it is intentional, not an omission. The rule table scopes RY002 to `if` conditions while RY001 and RY003 name `if` / `while`, and the `While` arm has never carried RY002 back to the first commit that introduced these rules. The helper preserves that split through `ConditionContext::allows_length_rule` rather than widening it. The binop scan was left alone: it checks RY032 after `infer_binop`, not RY100 after inferring a condition, so it is not the same shape.
- **Cached defuser set.** `check_lazy_default_reachability` rebuilt a `HashSet` from `DEFUSING_CALLS` plus every fn-table entry on every call, once per function per fixpoint iteration. The set now lives on the `Checker` behind an `Arc` and is built on first use. `collect_fns` clears it, which matches the existing invalidation story: collection is the only pass that writes `fn_table.fns`, and every table reset is immediately followed by a collection round.
- **One `ArgumentMatch` per call.** `argument_eval_mode` re-ran `match_arguments` per argument, making `infer_call`'s argument loop quadratic. `eval_mode_for_arg` consumes a match built once. `bound_argument_index_matched` does the same for the assertion path, which matched up to three times per call site. `argument_eval_mode` and `bound_argument_index` remain as self-matching wrappers for the call sites outside `infer/`.
- **`infer_data_frame`** calls `longest_arg_length` instead of repeating its combine ladder.
- **`infer_rep`** resolves `times` and `each` through one `rep_count` helper instead of writing the negative-literal and non-literal early returns out twice.
- **`list_binding_origin`** is inlined into its single caller.
- **Doc comment moved.** The `structure(x, class = ...)` comment sat on `infer_injected_call`, which handles typeshed injection. It now sits on `infer_structure_call`, the function it describes, and `infer_injected_call` got a doc comment of its own.

## Verification

- `cargo fmt -p ry-checker --check`
- `cargo test -p ry-checker` — full suite, 594 tests across 16 binaries, all green.


sims1253:
Addressed both reviews in bc916fa.

**Stale `trusted_defusers` cache.** I took the mutation-site option: both propagators drop the cache when they set `params[].quoting`/`defused`, and `seed_caller_visible_signatures` drops it when a seed wrote flags (`seed_caller_visible_signature` now reports whether it applied). A single clear at the top of `run_fixpoint_inner` would not guarantee a fresh pass 3: the cache is built mid-loop (iteration 1's `refine_fn_return` walks enter nested literals), the build is first-use-only, so later iterations reuse an iteration-1 snapshot through pass 3. Clearing at every writer makes the invariant local. The invariant comments in `misc.rs` and `lib.rs` now name the four invalidation sites.

**Test outcome.** Your exact divergence cannot fire through the public API today. In `first_executed_identifier` the trusted-defuser branch and the fallback branch both recurse into `func` only, so a bare `g(x)` never counts as a guaranteed force either way; only `base::`/`rlang::`-qualified `abort`/`stop`/`warning`/`message` force arguments, and those are not defusers. I verified this on the unfixed code: your scenario stays silent with the stale cache. The set has been inert for RY098 output since f2db1dd. So `defuser_cache_reflects_propagated_quoting` pins the invariant white-box instead: after the fixpoint, a cached set must contain the generic whose quoting arrived only via S3 propagation. It fails on the pre-fix code (the stale snapshot holds `g.foo` but not `g`) and passes with the fix.

**Second review.** Added the Unreleased → Changed entry for the once-per-call argument matching, the single function-entry walk, and the cached defuser set.

`cargo fmt`, `cargo clippy -p ry-checker --all-targets -- -D warnings`, and full `cargo test -p ry-checker` (461 lib tests) are green.

PR #146, opened 2026-08-31
refactor(cli): share check/dump plumbing, collect check args in one struct

sims1253:
## Summary

This PR is stacked on #133 and should merge after it.

`ry check` and `ry dump-types` ran near-identical copies of the same pipeline code. This PR moves that code into one place. CLI behavior does not change.

## Changes

- New module `crates/ry-cli/src/pipeline.rs` holds the shared front half of both commands.
- `pipeline::parse_files` replaces the two thread-local parser pools (`PARSER` in main.rs, `DUMP_PARSER` in dump.rs). One pool now serves both. Each command passes its error policy as a callback: check reports and counts every failure; dump-types warns and skips an unparseable file, and aborts on an unreadable one. The thread-local rationale comment moved with the pool.
- `pipeline::discover_config` replaces the duplicated `Ok(Some)`/`Ok(None)`/`Err` match in both commands. The two search-start rules stay separate and now say so in comments: check anchors at the first path's parent, dump-types at the first input itself.
- `pipeline::group_by_package_root` and `pipeline::workspace_context` replace the copied `BTreeMap` grouping loop and the six-field `WorkspaceContext` copy. `workspace_context` splits the resolved scope into the checker input and the degraded-scope notes; each command still reports those notes its own way. The helper lives in ry-cli; ry-workspace is untouched.
- `run_check` takes a `CheckArgs` struct and `run_check_once` a `CheckContext` struct. Both `#[allow(clippy::too_many_arguments)]` are gone. The file set stays a `run_check_once` parameter because watch iterations change it.
- `CheckArgs` now derives `Args` and backs `Cmd::Check`. The bare `ry` default is `CheckArgs::default()`. Flag defaults share one source: empty vecs, unset flags, and absent options default through their types; `--output-format` reads `DEFAULT_CHECK_OUTPUT_FORMAT`; `--color` and `--min-confidence` read the enums' `Default` impls.
- `CheckInput::files` drops the dead `version: i32` field. Both call sites passed literal `0`.
- Test module: one `check_files(paths, repo_root)` helper replaces ten copies of the same `run_check_once(...)` tail.

## Verification

- `ry --help`, `ry check --help`, `ry dump-types --help`, and the zsh/bash/fish completions are byte-identical before and after (captured and diffed).
- `cargo fmt -p ry-cli`
- `cargo clippy -p ry-cli --all-targets -- -D warnings`
- `cargo test -p ry-cli` — all suites pass (34 unit, 37 config e2e, 11 dump-types e2e, plus testkit/lsp/typeshed)
- `cargo check --workspace`

No changelog entry: nothing user-visible changed.


sims1253:
Addressed both review nitpicks in e4ff5a0:

- `pipeline::read_r_source` and `pipeline::enclosing_package_root` are now plain-private. Grep confirmed their only callers live in `pipeline.rs`; the `pub(crate)` was a leftover from when they sat in `main.rs`.
- Added `unreadable_discovered_file_aborts_with_single_stderr_line` to `dump_types_e2e.rs`. It pins the rewired abort path: a discovered file with mode 0o000 produces exactly one `ry: <path>: <error>` line on stderr, exit code 1, and no JSON on stdout. The test verifies the chmod actually denies the read first and skips when it does not (running as root), so it stays portable. The nonexistent-input case stays covered separately by the `!root.exists()` guard.

PR #148, opened 2026-08-31
refactor(checker): collapse the six collect.rs AST walkers into one

sims1253:
## Summary

- collect.rs carried six recursive AST walk families that differed only in a one-line leaf test: `stmt_captures_all_arguments` / `expr_captures_all_arguments`, `stmt_quotes_parameter` / `expr_quotes_parameter`, `stmt_captures_promise_parameter` / `expr_captures_promise_parameter`, and `bquote_references_parameter`.
- One generic walker pair now carries the traversal: `stmt_any` takes a predicate, `expr_any` applies it at every expression node. Each family reduces to its predicate plus one call.
- 253 lines of walkers became 96 lines of walker plus predicates.
- Stacked on #142 and merges after it.

## Changes

- Add `stmt_any` / `expr_any` in `crates/ry-checker/src/collect.rs`. The predicate runs at each expression node; the traversal recurses into compound statements, callees, and arguments. `Stmt::FunctionDef` and `Expr::Function` stay opaque, as before.
- Re-express the families as four leaf predicates: `captures_all_arguments` (match.call / sys.call), `quotes_parameter` (substitute, promise-capture helper, bquote template), `captures_promise_parameter` (promise-capture helper, dots capture), and `unquotes_parameter` (`.(param)` inside a bquote template).
- `parameter_is_quoted` calls the walker directly with those predicates.
- Leave the sibling walkers alone. `collect_parameter_uses_*` accumulate side effects, `first_parameter_use_*` stop at the first use, `block_force_flow` tracks fall-through. Different semantics, not the same shape.
- Second commit fixes one bug the diff surfaced; see below.

Arm-by-arm diff before unification: every `Stmt` arm and every `Expr` arm agrees across the families. Only the `Expr::Call` leaf test differs.

The one disagreement: `bquote_references_parameter`'s `Expr::Block` arm delegated to `stmt_quotes_parameter` rather than its own predicate. Decision: bug, fixed in the second commit. The delegation computed the same value the caller's structural recursion already computed, so it decided nothing; a deliberate no-op delegation makes no sense. Meanwhile it skipped the unquote scan, so `bquote({ 1 == .(x) })` lost quoting while `bquote(list(.(x)))` kept it. In R, `bquote` substitutes `.(x)` anywhere in the template, braces included. The fix walks braced bodies with the unquote predicate like every other arm. Failing test included: `bquote_dot_in_a_braced_body_still_quotes`. The first commit preserves the delegation verbatim, so the refactor and the fix are reviewable apart.

## Verification

- `cargo fmt -p ry-checker` clean.
- `cargo clippy -p ry-checker --all-targets -- -D warnings` clean.
- `cargo test -p ry-checker` green: 461 lib tests plus corpus snapshots, metamorphic, oracle, and invariants suites. The corpus snapshots pin diagnostics byte-for-byte; both commits passed them, so the refactor is behavior-preserving and the fix changes only the braced-bquote case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where unquoted expressions inside braced `bquote` bodies were incorrectly analyzed as eagerly evaluated.
  - Unbound names within expressions such as `.(x)` are now handled correctly, preventing erroneous RY010 (`unbound-variable`) diagnostics.
  - Added coverage to ensure quoting behavior remains correct for braced expression bodies.


sims1253:
Addressed both review points in a701681.

- Added the `### Fixed` bullet under `## [Unreleased]`: a `.(x)` in `bquote({ 1 == .(x) })` is now recognized as quoting, so an unbound name passed at the call site no longer gets RY010. The PR body already described the fix, so I left it as is.
- Applied your `stmt_any` wording verbatim and fixed `expr_any`'s header to match: both now say `pred` runs at call nodes, and `expr_any` notes the predicates early-return on non-call nodes, so that is equivalent to every node. I did not hoist `pred(expression) ||` above the match — it would run every predicate on every node for no gain, so the wording is the safer fix.

cargo fmt, clippy -D warnings, the 461 lib tests, and the corpus suite are all green locally.

sims1253:
Addressed the docstring-coverage warning: documented the five diff-touched functions that lacked docs (`parameter_has_normal_use`, `is_single_promise_capture`, `is_dots_promise_capture`, `bquote_references_parameter`, and the new braced-bquote test) in b7ac382. Docs state what each function does or what the test pins; pre-existing undocumented functions outside this diff are left as they were.

Issue #49, opened 2026-08-06
Migrate hardcoded function-name lists to typeshed stubs (stub data exists)


sims1253:
## Context

The pattern audit (`docs/plans/31-pattern-audit.md`) found several hardcoded function-name lists in ry-checker where the r-typeshed stubs already have the relevant data. These are the lists that can be migrated without schema changes — only code changes in the checker to consult the stubs.

Blocked on: each list's standalone function (`expression_has_list_origin`, `is_numeric_truthiness_idiom`, `DEFUSING_CALLS`) does not have access to `self.resolve_typeshed_sig()`. Converting them to Checker methods or threading the typeshed through the call chain is a non-trivial refactor.

### `expression_has_list_origin` (mod.rs:66, 3 names: list, lapply, Map)
- Stubs declare `return.mode: "list"` for all three.
- Migrate by consulting the resolved return type's mode instead of matching names.

### `is_numeric_truthiness_idiom` (index.rs:505, 5 names: length, nrow, ncol, NROW, NCOL)
- Stubs declare `return: { mode: "integer", length: "1" }` for all five.
- Migrate by consulting the resolved return type's mode+length.

### `DEFUSING_CALLS` (misc.rs:853, 14 names)
- 7 already have `eval: captures_promise` in stubs; 7 don't (needs r-typeshed stub completion: `expression`, `quote`, `substitute`, `bquote`, `alist` in base; `expr`, `exprs`, `quo` in rlang).
- The stub-driven `promise_capture_index()` (collect.rs:843) already builds a global index from the `eval` field. Once the 7 missing members get their `eval` fields in r-typeshed, `DEFUSING_CALLS` can be replaced by a lookup against the index.

## References

- Pattern audit: `docs/plans/31-pattern-audit.md` findings R2-R5
- Completed migration (example): `SCALAR_REDUCTIONS` → `is_typeshed_scalar_reduction()` (recall.rs)
- Related: #40 (schema fields for lists that need new fields), #41 (NSE/defusing consolidation)

sims1253:
Status update from the cleanup sprint (#160 did the in-repo half of this):

[...]

Issue #41, opened 2026-08-06
Consolidate three independent NSE/defusing function encodings


sims1253:
## Problem

Three independent encodings of overlapping NSE/defusing function knowledge exist in ry-checker, identified by the pattern audit (`docs/plans/31-pattern-audit.md`, finding D2):

| Encoding | Location | Members | Purpose |
|----------|----------|---------|---------|
| `DEFUSING_CALLS` (const) | `infer/misc.rs:853` | 14 | RY098 lazy-default analysis |
| `is_nse_symbol_fn` (match) | `infer/index.rs:597` | ~30 | Suppress RY010 on bare symbol args |
| `promise_capture_index()` (stub-driven) | `collect.rs:843` | dynamic (from stubs) | Quoting/defusing parameter analysis |

13 names appear in both `DEFUSING_CALLS` and `is_nse_symbol_fn`. Neither hardcoded list consults the stub-driven `promise_capture_index`. The lists can drift apart — adding a function to one without the other creates a silent inconsistency.

## Constraint

The stubs' `eval` field (`captures_promise`, `quoted_expression`) is the declarative source of truth, but only 7 of ~30 names in `is_nse_symbol_fn` have it. Completing the `eval` entries for the remaining members (base: `quote`, `substitute`, `bquote`, `expression`; rlang: `expr`, `exprs`, `quo`) is in the r-typeshed repo.

## Suggested approach

1. Complete `eval` field in r-typeshed for the 7 base/rlang members.
2. Replace `DEFUSING_CALLS` with a `promise_capture_index()` lookup + fallback for unstubbed names.
3. Audit `is_nse_symbol_fn` against the index; remove names the stubs cover, leaving only names for packages not in the typeshed (data.table, ggplot2).
4. Add a consistency test that checks every `is_nse_symbol_fn` member against the index and flags the overlap.

## Reference

- Pattern audit: `docs/plans/31-pattern-audit.md` finding D2

sims1253:
Concretely mapping the three encodings after the cleanup sprint (#160 centralized encoding A but did not merge it with B):

- **A — static list:** `semantic_lists::DEFUSING_CALLS` (14 names). Consumed by `build_trusted_defusers` (RY098 seeding) — now registered with a typeshed-agreement coherence test.
- **B — stub `eval` metadata:** `EvalMode::CapturesPromise` etc., consumed via `promise_capture_index()` (collect.rs) and `is_promise_capture`.
[...]

PR #149, opened 2026-08-31
fix(checker): honest NSE name lists with a stub-overlap guard

sims1253:
## Summary

`is_nse_symbol_fn` is the hardcoded half of the NSE knowledge. The typeshed `eval` metadata is the other half. This PR deletes the entries the stubs already cover, keeps the ones they do not, and adds a guard so the two halves cannot overlap silently. It also wires `generic_argument_may_dispatch` to the same `s3_generics` list the dispatch path reads. Stacked on #145 (which stacks on #137); merges after both.

## Changes

- Removed eight entries from `is_nse_symbol_fn`:
  - `tidyselect` is a package name, not a function. A call spelled `tidyselect(...)` now evaluates its arguments.
  - `alist` is covered by the base stub's `quoted_expression` dots. The stub path also keeps the list return type.
  - `enexpr`, `ensym`, `enquo`, `enquos`, `ensyms`, and `quos` are covered by rlang's `captures_promise` metadata. Loaded or qualified calls resolve the stub. Unqualified calls without `library(rlang)` now evaluate their arguments.
- Kept every member without stub coverage. `quote`, `substitute`, `bquote`, and `delayedAssign` have base stubs without `eval` fields. `makeActiveBinding`, the ggplot2 and data.table names, the remaining rlang names, and `all_vars`/`peek_vars` have no coverage. Blocked on r-typeshed.
- The list is now the `NSE_SYMBOL_FNS` const so tests can enumerate it.
- `generic_argument_may_dispatch` now consults `typeshed.globals.s3_generics`, the same list the S3 dispatch path reads. The stub list omits `round`, `mean`, `log`, `sqrt`, and `exp`, so a five-name fallback stays and a comment says why.
- `assertion_call_target` and the stub-assertion path in `infer_call` now carry cross-reference comments. Both sites encode the same knowledge. The fold-in is blocked on r-typeshed (#41).
- CHANGELOG gains a `### Fixed` entry: RY010 now fires where bogus or redundant entries suppressed it.

## Verification

- New guard `nse_symbol_fallback_does_not_overlap_stub_promise_capture` fails when a member is stub-covered and lists the overlap. Verified by temporarily re-adding `enquo`.
- New tests pin the stub-driven paths for `alist` and the six rlang helpers. They pass before and after the removals.
- New tests pin the two observable changes: a literal `tidyselect(...)` call and an unloaded unqualified defusing call now report RY010.
- `cargo fmt -p ry-checker`
- `cargo test -p ry-checker` — full suite green, corpus snapshots included.

Implements the in-repo part of #41 and #49


sims1253:
Addressed all three findings in c6dc2b1 (on top of a merge from the moved base).

**Guard widened.** `nse_symbol_fallback_does_not_overlap_stub_eval_modes` now walks the same stubs the old index builder walked (`load_base_cached()` plus every `known_packages()` entry) and records each function's non-Normal eval modes. A member fails when it carries `quoted_expression`, `captures_promise`, or `quoted_symbol`, and also when it carries `data_mask` or `tidy_select` unless it sits in `data_mask_exemptions` with a reason. The exemption list is empty. Mutation check: re-adding both `alist` and `all_vars` fails the guard with `["alist (QuotedExpression)", "all_vars (DataMask)"]`; the old guard saw neither. `promise_capture_index` went back to private since the guard no longer reads it.

**`all_vars` removed.** dplyr — the package it is called through — declares `expr: data_mask` for it, and the fallback intercepted even `dplyr::all_vars(...)` before signature resolution. Same trade as the rlang helpers. Pins: `dplyr::all_vars(x)` and `library(dplyr)` + `all_vars(x)` data-mask the expression (no RY010 on unknown columns, and a literal `RY040` inside the masked expression still fires, which the old fallback skipped), and an unloaded unqualified `all_vars(x)` evaluates its argument (RY010 on unbound names). Re-adding `all_vars` fails both pins. The membership-rule comment and the CHANGELOG `### Fixed` bullet now cover it. `peek_vars` stays: no vendored stub mentions it at all.

**Dispatch fallback aligned.** `generic_argument_may_dispatch` derives `round`/`log`/`sqrt`/`exp` from `s3_group_generic(&name).is_some()` and keeps only `mean` hardcoded. Inert today, as noted.

`cargo fmt`, `cargo clippy -p ry-checker --all-targets -- -D warnings`, and full `cargo test -p ry-checker` (468 lib tests) are green.

PR #153, opened 2026-09-01
ci: track pullfrog by its v0 tag per the vendor's pinning guidance

sims1253:
## Summary

This records a conscious pinning decision for the pullfrog action: take the
vendor-recommended moving `@v0` tag for this one action instead of the digest
pin #138 installed.

The reasoning, stated in full: pullfrog documents the action as a thin
bootstrap — the agent resolves from npm at runtime, so a ref pins only
`action.yml` and the `post:` cleanup step, and a SHA therefore buys little
reproducibility. The vendor's stale-cleanup warning applies to a bare SHA with
no updater; this repo runs weekly Dependabot on `github-actions`, so the
SHA + Dependabot combination the vendor endorses for security policies was
also available. This PR takes `@v0` because the wrapper pin adds a maintenance
step without adding reproducibility. The accepted trade-off is tag-move
exposure on the wrapper, bounded by the agent's token model (short-lived OIDC
installation tokens; the workflow's GITHUB_TOKEN is never consumed).

Review feedback from the first round is incorporated: the initial framing
stated the stale-cleanup warning unconditionally and omitted the Dependabot
context. Both are corrected here and in the zizmor rationale.

## Changes

- `.github/workflows/pullfrog.yml` — `uses: pullfrog/pullfrog@v0`
- `.github/zizmor.yml` — `unpinned-uses` ignore for pullfrog.yml recording the decision and its trade-off; artipacked comment corrected

## Verification

CI runs zizmor, yamllint, and actionlint on this change (local runs by the
first review round were clean). No behavior change beyond the ref; the
workflow is otherwise untouched.

Docs: https://docs.pullfrog.com/versioning


sims1253:
Both CodeRabbit comments handled:

- **Scoped exemption (implemented):** the filename-wide `unpinned-uses` ignore is gone. The ignore is now inline on the `uses:` line itself — `# zizmor: ignore[unpinned-uses]` — so it exempts only the `pullfrog/pullfrog` ref and cannot silently cover a future unpinned step in that file. The full decision record (thin bootstrap, SHA + Dependabot alternative, accepted trade-off) moved to a comment above the step, where a reader of the workflow finds it.
- **`persist-credentials: false` (skipped):** this exact question was adjudicated in #99 and re-verified in #138 — pullfrog's documented model mints its own short-lived OIDC installation tokens and never consumes the workflow's `GITHUB_TOKEN`, and the job token is `contents: read`, so the persisted checkout credential is inert. The file also carries the vendor's "DO NOT EDIT EXCEPT WHERE INDICATED" header, and the checkout shape is not in the indicated region. Changing it would deviate from the vendor-emitted reference shape for no real gain.

PR #155, opened 2026-09-01
test: consolidate copy-pasted checker tests into table-driven forms

sims1253:
Cleanup sprint, PR 3 of 7. Net **−405 lines** (12 files, +378/−783). No production behavior change; all changes are tests or `#[cfg(test)]` code.

## What

- **Table-driven conversions** (~30 folded test functions): the 20 pin-length literal-constructor tests, 6 pipe smoke tests, the `$`/`[[` dot-pronoun twins, both narrowing guard clusters (including one the audit missed), and 3 cross-file scope-resolution twins that differ only in a one-line fixture.
- **Single-run restructuring**: 9 NSE tests + the closure trio ran the checker *twice* on identical sources (`check(src)` then `check_with_scope(same src)`); they now run once and assert both diagnostics and scope from that run.
- **Strengthened instead of deleted**: `lapply_anon_callback_infers_integer` and `vapply_uses_fun_value_template` previously hedged with a bare "no RY010"; they now pin the inferred modes (List / Double) plus RY040 behavioral assertions. `long_else_if_force_flow_completes` now asserts the concrete (empty) diagnostic set.
- **Deleted 4 tautological tests**: `is_operator_generic`'s positive loop was `OPERATORS.contains` checked against `OPERATORS`; same shape for the FFI test; the testkit round-trip test was subsumed by the exact-length + back-to-back framing tests; the second VS Code e2e case asserted `diagnostics.length > 0` where the surviving case asserts the specific code on the same fixture.

## Coverage argument

Test functions went 966 → 930 (−36, exact accounting per group). An independent audit verified every folded row maps 1:1 to an old test (byte-identical R sources and expectations), via literal-multiset diff old-vs-new.

## Known cosmetic deltas (flagged in review, intentionally kept)

- Folded tables rename helper functions to `f` and normalize two fixture filenames to `R/a.R`/`R/b.R` — names are not load-bearing for these cases.
- Some diagnostic assertions now run through `check_with_scope` (which additionally surfaces RY000 parse errors); outcomes verified identical for these sources.

## Gates

`cargo test --workspace` (930 pass), clippy `-D warnings`, fmt, VS Code unit/lint/tsc/compile all green locally; e2e suite runs in CI (needs xvfb).

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for type inference, scope resolution, closures, pipes, data masking, and constructor lengths.
  * Added stronger assertions for inferred types, function returns, result scopes, and diagnostics.
  * Consolidated repetitive cases into clearer, table-driven test scenarios.
  * Refined cross-file and project-level resolution coverage.
  * Removed redundant test cases and imports without changing product behavior.

### DIFF
--- a/crates/ry-checker/src/infer/index.rs
+++ b/crates/ry-checker/src/infer/index.rs
@@ -770,26 +770,19 @@ pub(crate) fn assigned_names_in_body(body: &[Stmt]) -> HashSet<String> {
 #[cfg(test)]
 mod operator_generic_tests {
     use super::is_operator_generic;
-    use crate::semantic_lists::OPERATORS;
 
-    /// `is_operator_generic` must recognize exactly the members of
-    /// `semantic_lists::OPERATORS`: the same constant backs the S3
-    /// method-name splitter in `lib.rs`, so any divergence between the two
-    /// representations is an inconsistency (issue #42). The negative
-    /// samples pin the set to the Arith + Compare members -- Logic,
-    /// assignment, sequence, and access operators are operator symbols
-    /// for RY010 suppression but never operator generics, `%in%` is a
-    /// function-backed infix operator outside both dispatch groups, and a
-    /// full method name like `+.foo` is split before this predicate runs.
-    /// Reinstating a separate hardcoded symbol set here fails this test.
+    /// The negative samples pin `is_operator_generic` to the Arith +
+    /// Compare members of `semantic_lists::OPERATORS` (which the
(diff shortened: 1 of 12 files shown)


sims1253:
Addressed the codeRabbit path-matching finding in 10d557a: the three `ends_with("…/…")` string lookups now go through a `path_matches` helper that compares `Path` components, so they work on Windows separators too. Assertions untouched; `-p ry-checker` suite, clippy, and fmt green.

PR #156, opened 2026-09-02
refactor(lsp): comment diet and dedupe in backend.rs, drop stub-config parser

### LINKED ISSUE #150: Split ry-lsp backend.rs into modules
## What

`backend.rs` holds the whole server in one file: `State` and its caches, folder-config discovery, project checks, document sync, and diagnostic publishing. The seams already exist as comment sections.

## Why

A reviewer must hold five subsystems in mind to change one. The file is the largest in the repo.

## Proposed split

- `state.rs` — `State` plus the parse/tree/scope cache primitives.
- `folders.rs` — folder contexts, config discovery, filters, baselines, stubs.
- `project.rs` — `ProjectCache`, folder partitioning, the per-folder check step.
- `documents.rs` — `update_doc`, incremental edits, `parsed_file`, `scope_for`, span helpers.
- `publish.rs` — `publish_diagnostics`, `schedule_diagnostics`.

`backend.rs` keeps `Backend`, the `LanguageServer` impl, and the URI helpers.

## Acceptance

- No behavior change; `cargo test -p ry-lsp` green and unchanged.
- Each new module under roughly 600 lines.

sims1253:
Cleanup sprint, PR 2 of 7. Net **−327 lines** (backend.rs: 2362 → 2052; full-line comment density 25.4% → 17.8%). Zero behavior change except one disclosed delta below. **Closes #150** — instead of splitting backend.rs into modules (which adds import/re-export boilerplate and removes nothing), this PR removes the actual bloat: triplicated design essays, history narration, and copy-pasted branches. The crate is already modularized where it's cheap (hints/util/diagnostics/index); what remains coupled in backend.rs is genuinely coupled.

## Comment diet (−235 comment lines)

Each of these had one authoritative copy kept, the rest reduced to accurate one-line pointers:
- debounce/generation mechanism (was explained 3×) → now lives on `schedule_diagnostics`
- "a corrupt reload never silently clears the baseline" (was 4×) → lives on `rebuild_folder_context` + the `Err` contract
- the didChange/parse interleaving script (was 4× across backend.rs/lib.rs/protocol_contract.rs) → lives at the `maybe_pause` call site

Deleted: "This replaces the per-file reconstruction…", "Factored from longest-prefix ownership…", "Split out of parsed_file so…", line-by-line WHAT narration, an 8-line freshness proof → 4 lines. Kept: every load-bearing contract (abort-batch-on-invalid-range, UTF-16/surrogate rationale, version-stamped trees, lock-dropping/no-disk-I/O, AST/text atomicity).

## Code dedupes (−152 lines)

- `load_workspace_stubs` + its private `WorkspaceConfig` deleted — it was a second, stubs-only `ry.toml` parser duplicating `ry-config`, and a drift hazard (it only knew `typeshed`). Both call sites now go through the real root `Config` → `load_stubs_from_config`. `toml` dropped from ry-lsp deps.
- `parsed_file`: two arms that differed only in `Some(&tree)`/`None` → one `parse_with_tree(path, &text, old_tree.as_ref())` call.
- `apply_incremental_change`: two byte-identical fallback branches merged.
- `did_change_workspace_folders`: the same `starts_with` closure written 6× → one helper.
- `Backend { client, state }` manual reconstruction → `self.clone()`.

## Disclosed behavior delta (reviewed as the only one)

A root `ry.toml` that fails *full* `Config` parsing/validation previously still contributed `typeshed` stubs (the lenient `WorkspaceConfig` ignored everything else); it now degrades entirely to defaults with one warn. That is the more honest behavior — the old path silently half-applied an invalid config. Adjacent log-only changes: a corrupt root config now warns at initialize; the old per-file stub-config error messages are gone. No test asserted the old behavior (verified against both ry.toml test fixtures).

## Review

Independently audited before push: code-only surface maps 1:1 onto the items above; the stub-loader path-anchoring (relative dirs anchored at the workspace root) and warn-and-degrade semantics verified identical; `protocol_contract.rs` changes verified comment-only.

## Gates

`cargo test --workspace`, clippy `-D warnings`, fmt — all green (hermetic worktree-local target dir; note the protocol tests require `<workspace>/target/debug/ry`, which is how CI runs them).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1373,7 +1373,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "toml",
  "tower-lsp",
  "tracing",
 ]
--- a/crates/ry-lsp/Cargo.toml
+++ b/crates/ry-lsp/Cargo.toml
@@ -18,7 +18,6 @@ tokio = { workspace = true, features = ["rt", "io-std", "sync", "time"] }
 tracing = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
-toml = { workspace = true }
 
 [dev-dependencies]
 ry-testkit = { path = "../ry-testkit" }
--- a/crates/ry-lsp/src/backend.rs
+++ b/crates/ry-lsp/src/backend.rs
@@ -1,11 +1,9 @@
 //! LSP backend: `Backend`, `State`, the `LanguageServer` impl, and the
-//! document cache / debounce machinery (extracted from
-//! `lib.rs` so `lib.rs` is just module declarations + `run()`).
+//! document cache / debounce machinery.
 //!
 //! All request handlers read the cached parse (`State::parsed`) and the
-//! cached single-file scope (`State::scopes`) instead of re-parsing /
-//! re-checking on every request. Diagnostics are
-//! debounced via `schedule_diagnostics`.
+//! cached single-file scope (`State::scopes`); diagnostics are debounced
+//! via `schedule_diagnostics`.
 
 use crate::diagnostics::{
     diagnostic_to_lsp, diagnostic_to_lsp_with_source, make_ignore_action, make_ignore_file_action,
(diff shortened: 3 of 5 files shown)


sims1253:
Addressed pullfrog's finding in 16ad4a2: `invalid_root_rytoml_degrades_entirely_to_defaults` pins the disclosed delta. The probe document sits outside the workspace root (the only channel where root-config stubs are observable); a valid-root control session proves the RY001 probe is live, and the broken-root session asserts the session initializes, diagnostics publish, RY002 still fires at Warning severity, and the broken `typeshed` stubs did not half-apply.

Issue #80, opened 2026-08-12
Zed extension: implement real binary integrity verification


sims1253:
## Problem

The Zed extension does not verify the integrity of the `ry` server binary it
downloads from GitHub releases.

Until PR #79 this was worse than absent — it was falsely claimed. The download
path called:

```rust
let _has_sha256_asset = release.assets.iter().any(|a| a.name == sha256_asset_name);
match Self::verify_checksum(None, &release_details.downloaded_binary_path) { ... }
```

`verify_checksum` skips the comparison entirely when `expected_hash` is `None`
and returns `Ok(())`, so every download passed. The sidecar asset was located
and then discarded. Meanwhile the surrounding comments, the error string ("may
be corrupted or tampered with"), `extension.toml`, and the CHANGELOG all
asserted that verification happened.

PR #79 removed the false claim and the dead code. This issue tracks
implementing real verification.

## Why the obvious fix does not work

Releases publish `.sha256` sidecars for the **archive**:

```
ry-cli-x86_64-unknown-linux-gnu.tar.gz
ry-cli-x86_64-unknown-linux-gnu.tar.gz.sha256
```

The extension only ever holds the **extracted executable**, because
`zed::download_file(url, path, DownloadedFileType::GzipTar)` fetches and
extracts in one call — the archive bytes are never available to the extension.
Passing the sidecar digest to `verify_checksum` would compare a digest of the
archive against a hash of the executable and fail 100% of the time.

Contrast `release-vscode.yml`, which verifies correctly precisely because it
can separate the two steps:

```sh
curl -fL "${base_url}/${archive}" -o "${archive}"
curl -fL "${base_url}/${archive}.sha256" -o "${archive}.sha256"
sha256sum -c "${archive}.sha256"
# extract only after verifying
```

## Options considered

1. **Download twice** — fetch the archive as `Uncompressed`, hash it, compare,
   then call `download_file` again to extract. Rejected: you verify the bytes
   from request #1 and execute the bytes from request #2. A security control
   with a built-in TOCTOU bypass is no better than the current state. Also
   doubles install bandwidth.

2. **Extract inside the extension** — fetch via `zed::http_client::fetch`, hash
   exactly those bytes, unpack with `flate2` + `tar` (and `zip` on Windows).
   Correct, but adds three decompression dependencies and a hand-rolled
   extraction path on `wasm32-wasip2`. Hold in reserve.

3. **Publish a digest of the executable** (recommended) — see below.

## Proposed fix

[...]

PR #157, opened 2026-09-02
chore: sweep dead code, collapse PACKAGE_SPECS, repair garbled strings

sims1253:
Cleanup sprint, PR 1 of 7. Net **−593 lines** (34 files, +213/−806). Everything deleted was verified dead by workspace-wide grep before deletion; the few judgment calls are called out below.

## Dead API (ry-checker)

`Checker::apply_filter`, `Project::apply_filter` (every real caller uses `apply_filter_to_diagnostics`; the "kept so apply_filter can run after check()" doc comments were stale and are rewritten), `Diagnostic::with_confidence`, `filter_default_disabled` (its doc claimed the LSP uses it — false since #144), and the `packages.rs` backward-compat shim (nothing imports that path; the one internal user now imports from `ry_workspace::packages` directly).

**Legacy suppression parser** (`parse_suppressions`, `parse_ignore_comment`, `line_before_comment`, `has_file_suppression`, `filter_suppressed`, ~89 lines): production code reached it zero times — every real entry point uses the `_from_comments` lexical API; only its own tests kept it alive. The 16 affected tests are ported to `parse_suppressions_from_comments` via a `scan_comments` helper that reproduces the parser's `Comment` semantics (verified against `ast.rs`/`parser.rs`); all 29 tests in the module keep their behavioral intent.

## Dead code elsewhere

- **editors/zed**: the hand-rolled SHA-256 + FIPS test vectors (~110 lines) — its own comment admitted no production path calls it. The NOTE explaining the archive-digest problem stays; real verification is tracked in #80.
- **editors/code**: `environment.ts` (never-constructed placeholder), the `ry.addExecutableToTerminalPath` setting (forwarded by settings, applied by nothing — the checkbox did nothing), plus verified-dead helpers (`getProjectRoot`, `pathExists`, `resolveVariables`, `showLogs*` commands, `setCommand`, `MINIMUM_VERSION` alias).
- **ry-core/ry-typeshed/ry-workspace/ry-config**: `ClassVector::with_class` (copy-paste twin of `from_slice`), `FunctionSig::params`/`return_` accessors over public fields, unconstructible `TypeshedError::Json`, `Default for DiscoveryLimits`, never-read `TruncationReport.root`, unused derives/re-export, hidden duplicate `ExplainTypeshed` subcommand, `ecosystem/diff.sh`, `.yamllint.yml`.
- **`r_version` config key** (the one deliberate schema change): documented "Reserved for future use; accepted but currently ignored", read by nothing. Removed — **breaking** for `ry.toml` files that set it (now an unknown-field parse error). CHANGELOG entry added under `[Unreleased]`. Revert this hunk if you'd rather keep the forward-compat affordance.

## PACKAGE_SPECS

The 130-line struct-literal table → a 33-line `&[(&str, &str)]` tuple list. Programmatically verified: same 32 packages, same order; `PackageSpec` struct deleted.

## Garbled strings

- RY032: "…current R errors for vectors" → "…R errors at runtime for vector operands". The purrr vendor snapshot (one line) and the Posit corpus ledger (48 RY032 entries) were regenerated via the real `posit_messages.py update` path; both the PR fast tier and the scheduled full tier of the corpus check pass locally (negative control confirms the harness catches the old drift).
- `dump-types --format` error: "expected one of: json" (a list of one) → "only `json` is supported".
- `ry check` with no files: no longer prints Rust `Debug` of a `Vec<PathBuf>`.

## Gates

`cargo test --workspace` ×3, clippy `-D warnings`, fmt, VS Code lint/tsc/compile + unit tests, zed `cargo test --locked` — all green. VS Code e2e and the zed wasm build run in CI only (need xvfb / wasm target). Independently audited before push (see the ledger catch above).


sims1253:
Addressed pullfrog's CHANGELOG note in 803dd41: one `### Fixed` bullet under `[Unreleased]` covering the RY032 reword plus the two CLI message repairs.

PR #160, opened 2026-09-02
refactor(checker): consolidate hardcoded name lists into the semantic registry

sims1253:
Cleanup sprint, PR 5 of 7 (stacks on #159). Addresses the in-repo half of #49/#41. **Net +174 lines** (+208/−34) — the one PR in this sprint that grows the tree, deliberately: the `semantic_lists` registry pattern mandates per-list claim docs, registry entries, and coherence tests (~15-20 lines per list × 5), and that guard rail is the point. The call-site literals it deletes were only ~15 lines. If you'd rather keep the tree shrinking and drop the registry overhead, rejecting this PR alone is coherent — the analysis it produced for #49/#41 survives in the issue comments either way.

## What

Five hardcoded name lists move into the registry, each with a coherence test:
- `S3_MATH_GENERICS` (27) / `S3_SUMMARY_GENERICS` (7) — replace the match arm in `s3_group_generic`; name sets verified programmatically identical (34/34, order-preserving). `mean` stays a documented special case: it is a plain S3 generic (`mean.bar` catches `mean(z)`; `Summary.bar` does not), so it cannot join the Summary list.
- `GROUP_GENERICS` (`Ops`/`Math`/`Summary`/`matrixOps`) — replaces the predicate duplicated in collect.rs and infer/index.rs.
- `S7_OBJECT_CONSTRUCTORS` — replaces the lib.rs match; also fixes its doc, which described a deleted promise-capture helper (`match.call`/`sys.call`) instead of S7 constructors.
- `QUOTING_FORMS` (`~`/`expression`/`vars`) — replaces the duplicated list in `infer_call` and `is_aliasable_function`.

Coherence details worth review: `acosh`/`asinh`/`atanh` exist in the vendored base stub only as **ambient functions**, so the agreement-test helper now also consults `globals.ambient_functions`/`globals.s3_generics` (no current entry's outcome changes; module doc updated to match). The R oracle for quoting forms runs live (`~`/`expression` verified; `vars` when ggplot2 is installed); the S7 oracle mirrors the shiny skip pattern.

## What was deliberately NOT migrated (the real #49 finding)

`expression_has_list_origin` (3 names) and `is_numeric_truthiness_idiom` (5 names) are **not** blocked on typeshed plumbing — both are already `Checker` methods. They're blocked on semantics: the stubs cover **60** functions returning `list` and **7** returning integer-1 (the hardcoded five *plus `Position` and `nobs`*). Consulting stub return modes would change real behavior — `if (nobs(fit))` would stop firing RY003; `strsplit()` results would gain list-origin marking feeding `has_list_origin` consumers. Each list needs a deliberate widen-or-stay decision with fixture coverage, not a mechanical swap. Full analysis posted to #49/#41.

No vendored stub data was touched. **Refs #49, refs #41** (does not close either — the remainder needs r-typeshed data and/or semantic decisions).

## Gates

`cargo test --workspace` (969 pass, 19/19 semantic-list coherence), clippy `-D warnings`, fmt; snapshots byte-identical. Independently reviewed: set/order fidelity, `mean` byte-identical, ambient-widening impact, doc accuracy all verified programmatically.


sims1253:
Addressed the S7 export-list finding in 6272a44: the coherence probe now checks `getNamespaceExports("S7")` (what `S7::name` call syntax actually resolves through) instead of namespace membership, matching the claim wording. Skip gates unchanged; CI's oracle job (with S7 installed) exercises the new probe.

PR #175, opened 2026-09-02
refactor: collapse second-wave duplication across cli, config, checker, lsp, workspace

sims1253:
Second wave of dedupe from the cleanup review, all behavior-preserving.
Builds on #173 (stacked; merge that first). Net **-110 LOC** plus fewer
lock round-trips per publish and per parse.

## CLI: one resolution path for check and dump-types

`ry check` and `ry dump-types` copy-pasted the same ~45-line per-package
resolution front half — comments included — and had already drifted on
which fallback roots apply. Both now call `pipeline::resolve_groups`,
parameterized by the fallback roots (check: config root; dump-types:
`--project-root`, then config root), so the two commands cannot disagree
about library scoping. `run_check`'s hand-rolled `search_start` block
(12 lines re-implementing `Config::discover`'s dir/file rule) is gone.
`explain-rule`/`rule` share one `ExplainRuleArgs` flatten; help output
verified byte-identical.

## Tests live in the crates they test

The rbuildignore escape-parity test and eight pure discovery tests moved
from ry-cli's main.rs to ry-workspace's shared tests (calling
`discover_r_files` directly; the `collect_r_files` shim is deleted), and
the two baseline tests moved to ry-config beside the existing baseline
test. Tests that exercise CLI end-to-end behavior stay.

## ry-config

`Config::defaults()` and `Default::default()` were the same constructor
with two names; one definition remains. `merge_cli` took 10 positional
parameters behind a clippy allow, with six tests dominated by
`Vec::new(), Vec::new(), None, None, 0, 0` filler; it now takes a
`#[derive(Default)] CliOverrides` struct.

## checker infer/

- `match_params`: the `param_names().collect() + match_arguments`
  two-liner appeared ~10 times; one helper now serves all sites
  (`higher_order_argument_match` deleted). `check_typeshed_call_arguments`
  and `check_user_call_arguments` share their arity prefix via
  `check_call_arity` over a two-method `CallFormal` view.
- `infer_condition`: the RY103-before-infer / RY100-window quad appeared
  in `if`, `while`, and pipe's `if` expression; it lives in one place
  now, parameterized by the context noun.
- `try_assign_value`: the assignment guard triple (class attribute ->
  replacement target -> plain assign) is one function.
- `literal_callee_mode` delegates to `infer_literal_default`;
  `strip_negation` is shared instead of re-implemented; the double
  user-parameter lookup in `infer_argument_types` computes once;
  `callback_literal_return` uses `join_all`; the `is_numeric` twins
  hoisted; `is_zero/one_literal` merged; `collapse_c_mode` and
  `is_ffi_primitive` inlined at their sole call sites (the `.Internal`
  rationale moved to `FFI_PRIMITIVES`' docs).
- Ten dead `_span`/`_name` parameters removed (four more became dead as
  a consequence and went too).
- `check_with_scope` re-uses `emit_diagnostics` instead of inlining its
  walk loop.

## ry-lsp backend hygiene

`publish_diagnostics`: the redundant second lock folds into the snapshot
block; the `BTreeMap<String, _>` partition keyed by a lossy string with
a `"__root__"` magic sentinel becomes a `Vec<FolderPartition>` built
from the already-cloned contexts (root last, matching the old sort
position); the per-file publish loop no longer re-locks and re-resolves
the owning folder per diagnostic file — the partition's snapshot rides
along, which is also more consistent (what is published comes from the
same snapshot that was checked). `parsed_file` reads cache, text,
version, and old tree under one guard, cloning the tree only on the
miss path. Two single-use wrappers inlined.

## ry-workspace

`read_description_packages` was re-read and re-parsed from disk once or
twice per source file while every sibling lookup was cached per package
root; it now has the same `entry().or_insert_with()` cache.

## Verification

-D warnings`, `cargo test --workspace` (43 suites, 947 passed, 0
failed, oracle ran with R installed). Targeted re-runs locally: ry-cli
81, ry-config 33, ry-lsp lib 49, all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,22 @@ suppression actions — and fixes a parser panic plus several editor issues.
 
 ### Changed
 
+- **One shared front half for `ry check` and `ry dump-types`**: both
+  commands resolve their per-package groups, workspace contexts, and
+  checker inputs through one pipeline helper, so their file sets,
+  resolution roots, and degraded-scope notes cannot drift apart.
+  Diagnostics and dump output are unchanged.
+- **Consolidated duplicated helpers across the crates**: `ry rule` and
+  `ry explain rule` share one argument struct; `ry.toml` merging takes a
+  single `CliOverrides` value instead of ten positional flags; the
+  checker's argument matching, condition inference, and plain-assignment
+  binding each have one implementation; the language server partitions
+  open documents per folder once, carries the owning folder through
+  publication, and reads the parse cache under a single lock; workspace
+  resolution caches DESCRIPTION reads per package root. Tests for
+  workspace discovery, `.Rbuildignore` translation, and baselines moved
+  into the crates whose code they exercise. Behavior, diagnostics, and
+  inferred types are unchanged.
 - **`enable` is honored per folder**: a workspace folder whose settings
   set `enable: false` is skipped: the language server publishes no
   diagnostics and returns no inlay hints for it. The setting was
--- a/crates/ry-checker/src/higher_order.rs
+++ b/crates/ry-checker/src/higher_order.rs
@@ -1,17 +1,6 @@
 use super::*;
 use crate::infer::*;
 
-/// `HigherOrderSpec` argument indices name formal slots, never raw call
-/// positions. One ordinary R argument match drives callback, source,
-/// length, and template lookup throughout the higher-order path.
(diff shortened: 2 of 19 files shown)


sims1253:
Fair point — the non-member assertion was the only executable pin on the `.Internal` exclusion, and prose alone doesn't stop a future addition. Restored next to the const in bc12ee9 (worktree-verified, `cargo test -p ry-core internal_is_not` green).

Issue #171, opened 2026-09-02
LSP test suite carries subsumed convergence properties and a filter test that cannot detect its named bug


sims1253:
## What

Two spots in the ry-lsp test suite consume effort without buying the coverage their names promise:

1. **`w10_session_converges_to_fresh_server` + resolver self-check are subsumed.** `tests/session.rs:910-931`, `:992-1008`, `:1013-1205` re-run convergence that `pr_session_seeds` in `tests/session_state_machine.rs` already covers with a fuller operation alphabet (its header says as much: it "extends the session convergence property with the full operation alphabet"). The valuable unique pieces — the UTF-16 transcript test and the #81 regression — are separable. Est. -330 LOC if dropped.

2. **`many_files_flat_filter_construction` cannot detect the bug it is filed under.** `tests/protocol_contract.rs:1353-1472` creates 32 *identical* files and asserts file_00 and file_31 publish identical diagnostics — identical files under one folder config produce identical output whether filters are compiled per-file or per-folder, so the asserted property is determinism, not precomputation (#46). It also carries the heaviest synchronization machinery in the crate (30-second drain deadline with a per-message loop, justified by the #90 flake at `:1403-1412`).

## Proposal

- Split the unique tests out of the w10 block, then delete the subsumed property and self-check.
- Replace `many_files_flat_filter_construction` with a lean "N indexed files each publish RY090" test using the standard bounded-drain helper — or, if per-folder filter precomputation should actually be observable, make the files *differ* so output divergence is possible.

## Acceptance

- [ ] UTF-16 transcript and #81 regression tests survive
- [ ] No test whose name promises a property its assertions cannot check

PR #176, opened 2026-09-02
refactor: share test harnesses, diet narration comments, rename suppress.rs to resolve.rs

sims1253:
Third wave of the cleanup review: test-harness consolidation, a comment
diet restricted to an explicit list, and one module rename. Builds on
#173 and #175 (stacked; merge those first). Net **-307 LOC**, no
production behavior change (one test fixture's inert second line was
unified with the shared source-variant enum).

## Shared test harnesses

- **ry-lsp session tests**: `tests/harness/mod.rs` now holds the single
  copy of seven helpers that were duplicated verbatim between
  `session.rs` and `session_state_machine.rs` (`ClientSession`,
  `first_line_utf16_len`, `apply_incremental_edit`,
  `normalize_diagnostics`, `spawn_session`, `join_session`,
  `sync_barrier`) plus one shared source-variant type replacing the two
  parallel enums. The `w10` convergence property and resolver
  self-check are untouched — dropping them is a separate decision
  (#171).
- **ry-lsp protocol tests**: `tests/common/mod.rs` holds the single
  copy of `Published`, `workspace_root`, `ry_binary`, the CLI-JSON
  conversion, and `published_from_lsp` that protocol.rs and
  protocol_contract.rs each carried ("adapted from" per the old
  header). The three local `format!`-based `file_uri` builders are
  replaced by `ry_testkit::file_uri` (canonicalize + percent-encoding
  — the local copies broke on paths with spaces).
- **configuration_refresh.rs**: three spawn helpers collapsed to one
  parameterized on capabilities + initializationOptions (wire format
  preserved: init options omitted when null).
- **checker tests**: a `check_with(src, setup)` helper plus the
  existing `parse_file` collapse 25 copy-pasted parse+configure+check
  blocks; 13 sites keep their explicit shape because they need the
  scope, a Project, or mid-run introspection back. `project.rs`'s
  private `parse` twin is deleted. `scan_comments` calls the real
  parser instead of re-implementing comment collection with a "holds
  by luck" caveat. Three copy-pasted literal-family test pairs are
  table-driven; two truly redundant discovery-extension tests merged
  into the stronger one.

## Comment diet (the reviewed list, nothing else)

Deleted or rewrote ~60 lines of comments that narrated the next line or
argued at a reviewer: the T7b archaeology block, pipe-desugar narration,
the "no point attaching an empty signature" if-restatement, five
construct.rs functions' `//` blocks converted to concise `///` docs,
the Project re-export ergonomics comment, the clap-subcommand narration,
the stale `root_filter` `.first()` paragraph, duplicated republish
comments, and the util.rs extraction-history note. All R-semantics and
design-constraint comments were left in place.

## Rename

`suppress.rs` -> `resolve.rs`: the module contains typeshed/package
signature and value resolution (plus the emit helpers), while every
suppression feature lives in diagnostics.rs — the old name misled
reviews. Pure `git mv` + module declaration + header doc; no functions
moved, no logic changed.

## Verification

-D warnings`, `cargo test --workspace` (all suites green, 437 checker
lib tests, full LSP session/protocol/state-machine suites).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,17 @@ suppression actions — and fixes a parser panic plus several editor issues.
 
 ### Changed
 
+- **Shared test harnesses, leaner comments, `suppress.rs` renamed to
+  `resolve.rs`**: the checker's inline tests gained `check_with` (parse,
+  configure, check) and a shared `parse_file`, replacing copy-pasted
+  parser/checker scaffolding; the language server's session and protocol
+  tests each extract one copy of their spawn/normalize helpers
+  (`tests/harness/`, `tests/common/`); near-duplicate literal-pair tests
+  are table-driven. Narrating comments that restated the next line were
+  removed. Contributor-facing rename: `crates/ry-checker/src/suppress.rs`
+  is now `crates/ry-checker/src/resolve.rs` (same code; it holds the
+  typeshed/package signature and value resolution plus the checker's
+  emit helpers). No behavior change.
 - **One shared front half for `ry check` and `ry dump-types`**: both
   commands resolve their per-package groups, workspace contexts, and
   checker inputs through one pipeline helper, so their file sets,
--- a/crates/ry-checker/src/collect.rs
+++ b/crates/ry-checker/src/collect.rs
@@ -72,20 +72,8 @@ impl Checker {
                         let _ = self.record_fn(name.to_string(), params, body.clone());
                     }
                     self.collect_forwarded_calls(name, params, body);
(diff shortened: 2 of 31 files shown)


sims1253:
On the T7b aside: agreed the rationale is now git-history-only, deliberately — the note read as change archaeology rather than a constraint. The executable guard against a re-attempt is the corpus itself (the refinement was reverted after two rounds of corpus regressions, so a re-attempt fails the suite, not just the comment). If it ever gets re-attempted seriously, that is the moment to write the constraint down properly.

Issue #152, opened 2026-08-31
One shared AST walker for the remaining hand-rolled visitors


sims1253:
## What

PR #148 gave `collect.rs` one walker. Six hand-rolled `Stmt`/`Expr` walks remain in `infer/`:

- `vector_intent_parameters` (mod.rs)
- `assigned_names_in_body` (index.rs)
- `collect_forwarded_calls_in_stmts/expr` (misc.rs)
- `collect_executed_identifiers` and `collect_identifiers_in_stmt` (misc.rs)
- the `r6_rebound_members` walkers (call.rs)
- `collect_condition_assignment_names` (binop.rs)

`ry-workspace` adds two more twins: the dynamic-bindings walk (lib.rs) and the attachment walk (packages.rs). Each is a fresh match over the same enums, roughly 80 lines.

## Proposal

Add one child enumerator or visitor in `ry-core` (for example `for_each_child` pairs with default walks). Re-express the walkers on top of it.

## Caution

The traversals differ on purpose: some walk call arguments, some skip quoted spans. The shared walker must expose hooks for those differences, not erase them. Diff the arms family by family first, the way PR #148 did.

## Acceptance

- Roughly 300-400 lines deleted.
- Corpus snapshots green.
- Each converted walker keeps a test that pins its difference.

PR #177, opened 2026-09-02
refactor(checker): convert remaining collect.rs walkers onto the shared walker

### LINKED ISSUE #163: Convert the remaining collect.rs parameter-walkers onto the shared walker
Follow-up from #161 (the shared `ry_core::walk` core) — flagged in its review.

Two hand-rolled Stmt/Expr recursion pairs in `crates/ry-checker/src/collect.rs` were outside #152's listed scope and remain hand-rolled:

- `collect_parameter_uses_in_stmt` / `collect_parameter_uses_in_expr` (~line 759)
- `first_parameter_use_in_stmt` / `first_parameter_use_in_expr` (~line 885)

They likely fit the same per-child-rules exclusion that kept the RY098 force family hand-rolled in #161 (early-exit/branch semantics rather than subtree skips), but if any of their arms reduce cleanly to `Walk` policy knobs + callback, they should go onto the shared core — and per #152's acceptance criteria, each conversion must ship with a test pinning its skip policy. The #161 review also demonstrated why: both pre-push regressions there were in walker families that lacked pinning tests at the time.

sims1253:
Closes #163. Final PR of the cleanup stack (builds on #173, #175, #176;
merge those first).

## Converted (4 families onto `ry_core::walk`)

- **`collect_parameter_uses_in_stmt`/`_in_expr`** (the issue's primary
  target): every arm reduces to `Walk::ALL` + callback. The fn-shadowing
  skip and the `for` loop-variable rebinding are per-node walker
  decisions; the two parent-context rules (a plain `p <- value` binding
  never evaluates its LHS in R — complex targets like `arr[p] <- v` stay
  visited — and direct symbol arguments of defusing/`missing` calls) are
  decided at the parent and the classified leaf visits are suppressed by
  node identity (each node is visited at most once per walk; identities
  are only compared, never dereferenced). The two `_in_*` halves
  collapsed into one walker-driven function.
- **`collect_declared_globals_stmt`/`_in_expr`**: cleanest case —
  `Walk::ALL`, no skip rules, all work at the `Call` node; the expr half
  is deleted. Visit order identical to the hand-rolled version.
- **`collect_fns_stmt`** (If/For/While arms): statement-level walk with
  `Descend::Into` only for control-flow bodies; the Assign work moved
  verbatim into a factored helper. The old recursion's exact visit set
  and side-effect order are reproduced.
- **`collect_nested_fns_stmt`**: same statement-level policy; the
  `<outer>$<inner>` mangling stays as callback logic.

## Kept hand-rolled (documented, same judgment as the RY098 force family)

**`first_parameter_use_in_stmt`/`_in_expr`**: a first-use query in
evaluation order whose rules select individual children, not whole
subtrees — R evaluates the RHS of a complex assignment before the target
(R-lang "Complex assignments") while the walker descends target-first;
branch combination must walk BOTH `if` arms past their first hits while
the walker's `Break` halts everything; and the `for` rebinding check sits
strictly between iterator and body walks, where the walker has no hook.
A comment block at the family states this.

## Pinning tests (13 new, mutation-verified)

Per #152's acceptance rule every conversion ships skip-policy pins:
shadowing closures, plain-vs-complex assign targets, `for` binding and
iterator, defusing/`missing` argument classification, `$` fields and
control-test visits, declared-globals namespace-only recording, branch/
loop-body fn collection, per-level name mangling — plus four tests
pinning the KEPT family's evaluation-order semantics. A mutation
flipping `dollar_args` was caught by exactly the intended pin before
being reverted.

No new `Walk` knobs were needed; both policies were expressible with
existing ones. The pass-1 double-walk item from the #163 discussion is
deliberately not attempted here (pipeline change, stays on the issue).

Net: the four hand-rolled recursions shrank ~200 LOC; with the ~210
lines of pinning tests the diff is +535/-326.

## Verification

-D warnings`, `cargo test --workspace` (green, no fixture/snapshot
modified), `--test corpus` (5), `--test oracle` (11 passed; full R
matrix is the one ignored case, CI runs it), checker lib 450 including
the 13 new pins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,19 @@ suppression actions — and fixes a parser panic plus several editor issues.
 
 ### Changed
 
+- **Remaining `collect.rs` walkers on the shared walker**: the
+  parameter-use collector, the declared-globals scan, the
+  function-definition collection, and the nested-definition collection
+  now express their traversal through the shared `ry_core` walker
+  (`Walk::ALL` for the first two; a statement-level policy that skips
+  control tests for the definition walks) instead of four hand-rolled
+  Stmt/Expr recursions. The `first_parameter_use` family stays
+  hand-rolled: it answers a first-use query in evaluation order whose
+  rules select individual children (the value side of a complex
+  assignment before its target, both `if` branches past their first
+  hits, the `for` re-binding between iterator and body), not whole
+  subtrees. Every converted walker ships with a test pinning its skip
+  policy; diagnostics and inferred types are unchanged (#163).
 - **Shared test harnesses, leaner comments, `suppress.rs` renamed to
   `resolve.rs`**: the checker's inline tests gained `check_with` (parse,
   configure, check) and a shared `parse_file`, replacing copy-pasted
--- a/crates/ry-checker/src/collect.rs
+++ b/crates/ry-checker/src/collect.rs
@@ -1298,6 +1298,11 @@ mod collect_walker_tests {
         assert!(checker.fn_table.fns.is_empty());
         let checker = collect("out <- lapply(xs, function() 1)");
         assert!(!checker.fn_table.fns.contains_key("out"));
+        // A block-valued condition is the one place `control_tests: false`
+        // is load-bearing: with `Walk::ALL` the walker would reach the
+        // block's statements and run the assign collection on them.
+        let checker = collect("if ({ fn_in_cond <- function() 1 }) 1");
+        assert!(!checker.fn_table.fns.contains_key("fn_in_cond"));
     }
 
     /// Nested definitions record under the mangled `<outer>$<inner>`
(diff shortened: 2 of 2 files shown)


sims1253:
Both points addressed:

- **Pinning gap (inline):** correct — the block-valued-cond case is exactly where `control_tests: false` is load-bearing and nothing pinned it. Added to `fn_definitions_collected_from_branch_and_loop_bodies` in b549b8a (with a comment stating why that case is the load-bearing one). Full workspace suite green (957 passed).
- **Pass-1 double-walk tracker:** spun off to #178 before this PR closes #163, including the open question of whether the conversion already removed enough of the cost to defer it.

sims1253:
Correct. I re-ran your mutation myself (control_tests flipped to true: all 450 lib tests stay green), confirming the callback's Expr-node skip is the load-bearing rule and the knob only spares the walker the visits. Comment reworded in fa59806 to state the real mechanism — and that the pin fails if the Expr-skip is relaxed, which is the composite policy it actually guards.

Issue #131, opened 2026-08-31
Fixpoint re-refines every function every iteration — wake callers by changed slot instead


sims1253:
## Why

The single biggest performance lever in rustc's next-gen solver stabilization push was **not rescanning pending work**: record what each stalled item is waiting on, and wake it only when that changes. Their counter-example is instructive — a fulfillment loop that drained and rescanned everything turned generated code with ~4000 obligations from 1.76s (old solver) into 46.85s (new solver) before the wake-by-dependency fix (rust-lang/rust#159933).

ry's fixpoint loop has the same rescan shape, at three spots:

1. `run_fixpoint_inner` (`crates/ry-checker/src/lib.rs:976-996`) runs up to `MAX_FIXPOINT_DEPTH = 8` iterations, and **each iteration** clones the entire `ReturnSlots` table (`let before = (*self.return_slots).clone()`), re-refines *every* in-scope function, then detects convergence by comparing whole vectors.
2. `propagate_s3_generic_quoting` (`lib.rs:1006`) scans all functions for matching `generic.class` names — O(fns × fns) — on **every** iteration.
3. The reverse call graph that does exist is **file-granular**: `compute_fixpoint_scope` (`crates/ry-checker/src/project.rs:511-517`) marks *every* function defined in any file that calls an affected function as affected.

## Proposal

- `refine_fn_return` reports **which return slots changed** in this pass (it already knows the slot id).
- Iteration N+1 refines only functions that call a function whose slot changed in iteration N — plus the S3/quoting propagation analogues, driven the same way.
- Collect caller sets **per function** during pass 1 (`call_sites` is per-file today). This de-coarsens both the in-loop wake set and the between-checks incremental dirty set in `compute_fixpoint_scope`.
- Index S3 methods by generic prefix once per pass-1 collection instead of scanning per iteration.

This is a **de-slopping** change as much as a perf one: it deletes the whole-table clone/compare and the per-iteration O(fns²) scan in favor of one per-function edge set, which is also the structure the incremental path already wants.

## Acceptance

- [ ] No whole-table clone or vector-equality comparison per iteration
- [ ] Iteration N+1 refines a subset: callers of changed slots, with a test proving a stable callee is not re-refined
- [ ] S3 propagation is no longer O(fns × fns) per iteration
- [ ] Metamorphic determinism suite (R2 serial/parallel, R3) green; all corpus fixtures + oracle green
[...]

PR #256, opened 2026-09-07
perf(checker): retain per-function refinement dependencies

sims1253:
Edits currently refine every function in a caller's file. Retain observed per-function callable reads across checks and combine them with forwarding and S3 metadata dependencies, so unrelated functions can keep their converged state. File-level diagnostic invalidation remains conservative.

Newly discovered alias attachments restart refinement from fresh collected tables; subsequent checks remain full while those attachments are absent from collection. Changes to non-slot callable bindings also force full refinement and diagnostic emission, including when function returns stay unchanged. Tests compare cold and incremental returns, complete caller-visible signatures, and diagnostics across repeated edits, aliases, callbacks, recursive calls, missing callables, metadata changes, divergence probes, and attachment removal.

The synthetic fixture visits only the edited leaf and its caller, skipping 101 unrelated functions. Criterion estimates on Ryzen 5 PRO 5650U, Rust 1.98.1, `RAYON_NUM_THREADS=2` (20 samples; baseline d688f77; separately built executables):

| Workload | Before | After |
| --- | ---: | ---: |
| Sparse caller edit | 886 µs | 811 µs |
| Glue leaf edit | 3.741 ms | 3.607 ms |
| Glue dependent edit | 9.774 ms | 9.517 ms |
| Glue LSP edit simulation | 10.320 ms | 9.621 ms |
| Glue cold check | 10.742 ms | 10.500 ms |

Cold and dependent-edit differences fall within Criterion's noise threshold. These are local measurements, not general speedup claims.

Validation: workspace tests, all-target clippy with warnings denied, formatting, and all 17 R oracle tests passed. All gates also pass on final head 2896cdc, including the callable-binding regression.

Addresses the per-function refinement portion of #131; dirty-file seeding remains file-wide.


sims1253:
@pullfrog review

Please review exact head 2896cdc. The callable_vars fix now snapshots and resets that set, forces full unseeded refinement when it changes, and independently forces every-file diagnostic emission. Regressions cover repeated changes, a constant-return function in an independent file, and top-level calls. Removing the emission gate alone reproduces stale RY070. All workspace tests, all-target Clippy, formatting, and 17 R oracle tests pass on this head.

PR #257, opened 2026-09-07
fix(checker): respect base structure argument binding

sims1253:
`structure()` treated `x=` as its payload, could replace an opaque payload with a later argument, and skipped class-argument evaluation. Qualified base calls lost class information. Match the real `.Data` formal, force it before attributes, and preserve its list schema when adding a known class. Name replacements discard stale columns.

Use the constructor model only for resolved base calls. Respect custom functions and base stubs; keep uncertain namespace/search-path calls opaque without forcing potentially quoted arguments. Class-vector parsing also requires base `c`, and remains conservative for named controls, duplicate or missing arguments, and unknown ellipsis expansion. Literal class removal and factor storage conversion follow R.

Required formals now retain their parameter marker during function-signature capture, so parameters named `structure` or `c` cannot accidentally gain base-constructor facts. Extract the existing literal-binding guard into the resolver and reuse it for namespace syntax checks without changing lazy-default behavior.

Validation: twelve focused constructor tests, workspace tests, Clippy, formatting, and the complete R oracle gate (17/17) pass on merged source head 48c772c. Final 6f9ba33 only updates the Posit message baseline: the same four previously reviewed rlang RY061 false positives were removed from both ledgers. The runtime witness and a regression that fails before the fix cover string call heads that resemble namespace syntax.


sims1253:
@pullfrog review

Please review final head 4a05b3a, including constructor provenance guards and the string-call-head regression. The earlier cb71e32 fixture failure is fixed. Twelve focused constructor tests and the runtime witness pass; final workspace gates are running.

PR #271, opened 2026-09-07
fix(parser): preserve omitted call arguments

sims1253:
The parser dropped unnamed empty arguments and discarded names on empty named arguments. For example, `f(, x)` became `f(x)`, shifting formal matching and sometimes borrowing the wrong payload type or forcing contract. Indexing used a separate source-text workaround that could not preserve all empty positions.

Preserve every actual slot with an explicit `Expr::Missing(Span)`, retaining names such as `f(a=)`. Calls and indexes now use the same comma-aware lowering. Missing values infer as unknown, and existing constructor/forcing guards reject them. Required-argument and printf arity reporting distinguish an occupied slot from a supplied value; optional defaults remain valid. Switch selection is unchanged.

Rust API change: downstream exhaustive matches on `ry_core::ast::Expr` must handle `Missing` separately from unsupported `Unknown` forms.

Validation: parser tests cover leading/interior/trailing holes, named omissions, comments, nested calls and comma-containing strings; checker tests pin positional matching, missing payloads, constructor guards, required/optional arguments, forcing arity and indexed column types. Live R witnesses cover defaults, required omissions and indexing. Final checker source `2a6e468` (with documentation follow-up `dd77cbf`) passed the full workspace, clippy, formatting, and full R oracle (17 tests). Before the final printf count amendment, comparing the integrated release binary with the exact #266 parent `71a2ae0` with `RY_NO_INSTALLED_LIBRARIES=1`, immutable copied binaries and fresh isolated git-archive copies of all 62 pinned Posit sources produced identical full JSON diagnostics for all 124 package-root and R/ checks. Earlier standalone ecosystem checks also passed, but shared-cache overlap makes the isolated comparison and CI the stronger corpus evidence.

#266 is merged; its constructor/operator changes are prerequisites, not part of this PR.


sims1253:
@pullfrog please review the parser missing-argument preservation and conservative consumer handling. This is stacked on #266; combined local gates are running.

sims1253:
Addressed the printf count regression in 2a6e468: omitted actuals no longer satisfy ordinary literal-format conversion counts. Nine focused missing-argument tests pass, including leading/trailing holes and gettextf. Positional `$` references and `*` widths retain the existing unsupported-format boundary. The one-conversion `sprintf("%d", 1L, )` case was already quiet before this PR and remains a separate missing-extra-slot gap; the two-conversion trailing-hole case now warns correctly. Full gates are running. Also clarified the RY091 rule description for bound-but-omitted values.

sims1253:
@pullfrog please review the printf count fix at 2a6e468 and RY091 description at dd77cbf. Positional/star formats remain outside the existing count-only contract; dedicated controls pin that boundary. Local full gates are rerunning.

sims1253:
Updated the RY091 registry summary in the final follow-up to match docs/rules.md. The nine rule-evidence checks pass; actual `ry explain rule RY091` output contains the same revised description. Formatting and diff checks pass. No inference behavior changed.

PR #275, opened 2026-09-07
Wait for workspace context before initial editor diagnostics

sims1253:
Opening a file while the initial workspace scan is running can publish RY010 for a configured environment binding before that binding reaches the checker. Initialization never refreshed that publication, so the false warning could remain until the next edit.

Defer diagnostic checks until the first current scan installs its workspace context, then schedule the open documents again. A superseded scan cannot release readiness or replay removed-folder documents. Removing every folder, or completing an empty or failed scan, also releases the startup wait.

Validation: the controlled first-publication regression fails on unchanged main `4eb045b` (test commit `e8caa75`) and passes with the fix. A second controlled test removes all folders while the original scan is held and checks that an outside document still receives diagnostics. All four input-boundary tests and the full `ry-lsp` suite pass; formatting and diff checks pass. Clippy passes with warnings denied. The scheduling barriers compile only with the existing test-util feature.

The 1,000-seed nightly LSP session property also passes on head `4c3f992` (696 seconds). All regular GitHub CI checks at that head, including both ecosystem corpora, passed.

Pullfrog reviewed the startup implementation at `142bf4d` with no required changes and independently verified both regressions fail when their respective fixes are removed. Initial diagnostics depend on the workspace scan completing or failing; a filesystem operation that never returns still delays publication. This is the intended context-readiness contract.

Head `3f9e4f9` integrates merged checker and parser fixes through #281 and resolves adjacent changelog entries. The startup entry now sits beside the editor fixes. The LSP implementation is unchanged from the reviewed implementation except for the comment documenting republish ownership. The full LSP suite passed after #277 integration; all CI passed at `0f4b4a5`, and final CI checks the combined branch.


sims1253:
@pullfrog please review current head 142bf4dfd26da94ba6b7ab170871ea6c42364a27. Earlier runs produced no review because the provider quota was exhausted. All regular CI passes; the full LSP suite and 1,000-seed nightly property pass on the same implementation.

PR #278, opened 2026-09-07
fix(checker): select proved literal switch alternatives

sims1253:
Base `switch` now infers only the selected alternative when its identity, literal scalar selector, and preceding environment are proved. Character matching preserves missing-alternative fallthrough and defaults; numeric/logical selection preserves truncation and out-of-range `NULL`. Selected assignments reach the caller, while unselected errors and writes are skipped.

The proof is deliberately bounded to an ordinary clean prefix. Masks, uncertain callees, namespace/syntax rebinding, ambiguous arguments, and uncertain selected-branch effects discard the result and caller facts. Whole base `list`/`structure` calls with inert literal actuals retain their existing constructor model. Known custom functions keep the ordinary call path; its eager-argument limitation has a live-R known-gap fixture. Bare dynamic selectors retain the existing all-alternative model, and vector-selector diagnostics remain out of scope.

Depends on #274 for correct decoded literal values. Parser #271 supplies explicit missing arguments. Tests cover selector binding, fallthrough/defaults, skipped and selected writes/errors, caller effects, masks, namespace rebinding, literal escapes, and implicit index dispatch.

Validation: full workspace tests, Clippy with warnings denied, formatting, default oracle (16 tests), and complete live-R oracle (17 tests) pass at `61e3423`. Thirteen focused tests pin the selection and proof boundaries. Diagnostics produced inside an effectful selected branch still use the existing inference rules; this does not solve general intra-branch effect analysis.

An immutable-binary comparison against parent `dad9b8d` (same parser/decoder inference prerequisites) checked 124 pinned package-root/R outputs for the 62-package corpus: 31 root findings disappear, none appear, and all 43 required true positives remain. These are documented uncertainty/coverage changes, not 31 individually proved bug fixes. The default 32-package comparison checked 64 outputs and removes six overlapping RY010 findings in cli/rlang, with no additions or owned-ledger changes. Both report sets regenerated and reconciled successfully, and both strict `--check` runs passed with the immutable final binary.

Parent refresh `22ea388` merges #274’s current base. Its only production delta from the validated source is the behavior-neutral sparse capture-index allocation cleanup from #273; full checker tests, all-target Clippy, and formatting pass after the merge. Corpus measurements above retain their stated exact binary provenance.

Latest parent integration `5bbf16b` includes reviewed decoder parent `4bb36c2` and the merged deferred/printf guards. All 614 checker unit tests and five corpus tests pass. Integrated workspace, workspace Clippy, formatting, and full live-R oracle (17 tests) pass. Refreshed CI corpus checks are running; earlier standalone corpus evidence above predates the inherited escaped-space amendment.


sims1253:
@pullfrog Please review the literal-selector proof, selected-branch effect boundary, and retained required corpus findings. Full code and R gates passed; strict report snapshot checks are still running.

sims1253:
@pullfrog Please review current head 22ea388. The previous run failed only because the provider quota was exhausted and produced no review. All regular CI checks and local gates pass; there is no newer active review run.

PR #279, opened 2026-09-07
Check printf arity only for proven base formatters

sims1253:
Custom `sprintf` and `gettextf` functions were checked against base format-string arity. For example, `sprintf <- function(...) 1L; sprintf('%d')` succeeds in R but produced RY094.

Run that diagnostic only when the shared callee classification proves the base formatter. Preserve genuine base arity warnings, custom/stub return contracts, and existing call inference and effects. Spelling aliases remain unproved. The editor example now uses explicit `base::sprintf` so its intended arity warning remains visible.

Validation: custom/formal/alias/package/namespace controls, base positive controls, stub contracts, live R witnesses, and the editor playground. Final workspace tests, clippy, formatting, full R oracle and semantic-list registry tests pass. Strict default and all 62 Posit package checks pass with unchanged reports and all required findings retained. No corpus ledger changes.

Stacked on #277 for the shared callee classification helper.


sims1253:
@pullfrog Please review the current head 8cd764a0f04d5f4dde13e5c59cbb7e1e79e1f5d2 against its parent PR #277. The previous run failed solely because the provider quota was exhausted, before producing a review. All CI and local gates pass; no newer review is active.

PR #280, opened 2026-09-07
fix(checker): match captured actuals to helper formals

sims1253:
Collect wrapper evaluation modes from the actual argument matched to each `captures_promise` formal. Previously, any named capture formal made the first actual look captured, and two other collection paths treated every actual as defused. For example, `delayedAssign("held", p)` should capture `p`, while `delayedAssign(p, 1L)` evaluates it as the target name.

The four collection consumers now share formal-specific classification, including named reordering, partial matching, and qualified base helpers. Normal environment arguments and explicit rlang controls stay separate from captured expressions. A normal use in another actual dominates capture without assuming lexical argument order is force order. The sparse inventory introduced by #273 remains sparse.

Vendors the complete base helper formals from [r-typeshed #25](https://github.com/sims1253/r-typeshed/pull/25), commit f87bf6c. Merge that reviewed prerequisite first.

Boundaries:

- Bare names still use the existing embedded-inventory union; collection does not resolve lexical callee identity or read user stubs.
- Forwarded dots retain the existing blanket capture approximation. Runtime names may bind normal controls; this change handles explicit controls without introducing false diagnostics for ordinary forwarding wrappers.
- Unknown dots expansion preserves exact-named capture and signatures whose every formal captures. Mixed positional/partial occupancy stays conservative.
- Escaped names stay opaque for mixed signatures. All-capture signatures retain their capture fact without validating the name; existing RY090/RY091 diagnostics remain unchanged.

Validation on final d32ef7a: full workspace tests (including 596 checker library tests), Clippy with warnings denied, formatting, all 17 R oracle tests, and both default and full 62-package Posit corpus checks passed. The two pinned corpus checks ran serially in an independent cache and produced zero report drift; Posit retained all 687 readable message identities. Live R witnesses cover capture versus normal environment/control forcing, exact/partial/leading-dots matching, and escaped argument names.

Cost check against main d9902bc, using fresh hermetic processes and two Rayon workers: 200 capture wrappers added 1.05% instructions and 1.38% allocated bytes. Pinned purrr had negligible instruction change (-0.08%) and +0.05% allocated bytes. Diagnostic JSON was byte-identical across all profile runs; these are instruction/allocation measurements, not wall-time claims.

Hold merge until final checks and independent review complete. Pullfrog currently reports an external provider quota limit; no provider configuration has been changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Match promise-capture helpers by formal argument when collecting wrapper
+  evaluation modes. Recognize qualified base helpers and keep their environment
+  arguments, and explicit rlang controls, separate from captured expressions.
+
 - Preserve omitted call arguments and their names without shifting later
   arguments. Calls and indexes now share missing-position handling. The public
   `ry-core` AST adds `Expr::Missing(Span)`; consumers with exhaustive expression
--- a/crates/ry-checker/src/collect.rs
+++ b/crates/ry-checker/src/collect.rs
@@ -586,14 +586,28 @@ fn capture_signature_arguments(signature: &FunctionSig, args: &[Arg], captured: 
         .iter()
         .map(|a| a.name.as_deref().map(semantic_argument_name))
         .collect();
-    let matched = match_argument_names(&params, names.iter().copied());
+    // A forwarded dots expression expands to zero or more actuals, not one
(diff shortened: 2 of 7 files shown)


sims1253:
@pullfrog Please review final head d32ef7a21f800e6df2d575b1a47a708459cfa4fd. The previous run produced no review because the provider quota was exhausted; there is no active newer review. All local gates and CI pass. Review formal matching, forwarded-dots boundaries, and escaped all-capture handling; do not change provider configuration or merge.

PR #284, opened 2026-09-07
perf(infer): reduce expression-if scope allocation

sims1253:
Expression-form `if` previously kept both cloned scopes alive through both arms and cloned an else scope even when no else existed. Infer each arm in a temporary child, retain its type/effect flags, and release the child before its sibling. Both arms still use the original parent, and effects propagate after both finish.

On 1,024 bindings with 24 nested expressions, DHAT allocations fall 33.52% without an else. Peak live heap falls about 31.9% with or without an explicit else. The 461-file corpus saves only 0.037% of allocated bytes; peak stays unchanged. Diagnostics match byte for byte. These are allocation measurements, not wall-time or RSS claims. Exact totals, hashes, and reproduction are in `docs/corpus/if-expression-scopes.md` and its JSON record.

Adds regression tests for sibling isolation, NULL joins, and effect propagation, plus four Criterion inputs. A separate prerequisite commit simplifies two equivalent walker conditions so the existing code passes current Clippy.

Validation: `cargo test --workspace`; `cargo clippy --workspace --all-targets -- -D warnings`; `cargo fmt --all -- --check`; full R oracle matrix (17 tests); Criterion benchmark smoke run (four inputs).

Related to #130. This is a bounded improvement: scopes still clone, so the O(1)-snapshot acceptance criterion is unmet and #130 remains open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type and effect inference for conditional expressions, including more reliable handling of sibling branches and omitted `else` clauses.
  * Prevented bindings from one conditional branch from affecting another.

* **Performance**
  * Reduced temporary scope allocation during conditional-expression checking.

* **Documentation**
  * Added memory-profiling results and reproducible benchmarking guidance for conditional-expression checking.

### DIFF
--- a/crates/ry-checker/benches/performance.rs
+++ b/crates/ry-checker/benches/performance.rs
@@ -151,6 +151,37 @@ fn check_branch_scopes(c: &mut Criterion) {
     group.finish();
 }
 
+/// Expression arms discard bindings and need only one live child scope at a time.
+fn check_if_expression_scopes(c: &mut Criterion) {
+    let mut group = c.benchmark_group("check_if_expression_scopes");
+    for bindings in [128, 1024] {
+        for with_else in [false, true] {
+            let mut source = String::from("f <- function(flag) {\n");
+            for i in 0..bindings {
+                source.push_str(&format!("x{i} <- {i}L\n"));
+            }
+            source.push_str("result <- ");
+            for _ in 0..24 {
+                source.push_str("if (flag) (");
+            }
+            source.push_str("x0");
+            for _ in 0..24 {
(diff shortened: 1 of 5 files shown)


sims1253:
@pullfrog Please review current head 7f8f0f2bb75001c065ce2dddd5499d8a5cb5b2f0. The previous run stopped on provider request-rate limiting before producing a review. Local audit workers are paused to free capacity. The production change releases expression-if branch scopes sequentially; the report includes allocation and peak-heap measurements. All completed CI checks pass.

PR #286, opened 2026-09-07
fix(checker): allow Math and Summary member fallback

sims1253:
Calls such as `sum(structure(1:2, class = "widget"))` and `abs()` wrongly emit RY050 when an unrelated local Summary/Math method exists. These built-ins have default behavior without a class-specific method. Keep the existing unknown result on this path and suppress the false missing-method warning.

Tests cover every modeled Math/Summary member, retain the direct-generic missing-method control, and run ten fallback assertions in an R oracle. The initial test failed with RY050; R ran the fixture successfully.

Workspace tests, Clippy with warnings denied, formatting, and the full R oracle matrix pass. The full Posit and tidyverse corpus checks pass unchanged. A final comparison across 500 packages removes exactly one diagnostic: the false RY050 on `sum(freq)` in Hmisc `R/describe.s:550`, where `freq` retains its table class. No diagnostics were added or changed elsewhere.

Includes the same two walker boolean simplifications as #284 to pass Clippy on Rust 1.96.1. No behavior change is intended in those conditions.

R documents these members and group dispatch in its [group generic reference](https://stat.ethz.ch/R-manual/R-devel/library/base/html/groupGeneric.html); the executable oracle confirms their fallback behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,10 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Math and Summary member calls no longer emit RY050 just because an unrelated
+  class has a local group method. Built-ins such as `sum()` and `abs()` can
+  use their default behavior without a class-specific method.
+
 - Report an explicit nesting-limit error before deeply nested syntax can
   overflow the parser stack. The limit is 128 tree-sitter syntax levels.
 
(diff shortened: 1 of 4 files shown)


sims1253:
@pullfrog Please review current head b9653b2. The previous review stopped on provider request-rate limiting without findings. Local pi workers are paused. All CI checks pass; local full workspace, Clippy, formatting, complete R oracle, and both committed corpus gates pass. The final 500-package comparison removes exactly the Hmisc RY050 motivating this patch and changes no other diagnostic. Please use a single review agent if possible to keep provider concurrency low.

PR #287, opened 2026-09-07
fix(cli): render empty reports when no R files are discovered

sims1253:
Checking a directory with no discovered R files exits successfully but leaves stdout empty even with `--output-format json`. This breaks JSON consumers auditing data-only packages such as BH and carData.

Render the empty diagnostic list before the early return. JSON and GitLab now produce `[]`; JUnit produces an empty report. Human and GitHub stdout remain empty, and the existing stderr message and successful exit are preserved.

Real-process CLI tests cover all six formats and config-driven JSON when every source file is excluded. BH and carData both reproduce empty stdout on d5414ff and produce `[]` with this change.

Validation passed in CONTRIBUTING order: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --all -- --check`. The complete R oracle matrix also passes: `cargo test -p ry-checker --test oracle -- --include-ignored` (17 passed).

Includes the shared walker-condition Clippy cleanup prerequisite so the required gate remains clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,9 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- `ry check` now emits an empty JSON/GitLab array or JUnit report when no R
+  files are discovered, including when configuration excludes every source.
+
 - Report an explicit nesting-limit error before deeply nested syntax can
   overflow the parser stack. The limit is 128 tree-sitter syntax levels.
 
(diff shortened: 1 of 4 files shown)


sims1253:
Added the missing Unreleased / Fixed changelog entry in a873fb7. The entry covers both empty directories and configurations that exclude every source. This follow-up changes documentation only; the implementation passed workspace tests, Clippy, formatting, and the full 17-test R oracle matrix.

PR #290, opened 2026-09-07
fix(checker): respect scalar parameter guard result lengths

sims1253:
`is.null(x) || "value" %in% x` received RY032 because the checker treated the lookup table as determining membership-result length. Membership now tracks its left operand. Exact `length(x) == 1 && ...` guards also stop triggering the parameter-vector heuristic when the call resolves to base R; shadowed, parameter-bound, unrelated-namespace and uncertain search-path calls keep their previous behavior.

The frozen 500-package comparison removes 10 RY032 diagnostics across seven packages and adds none. Each removed expression uses the parameter as the membership lookup table. The broader candidate removed 78, but its exact-length exemption admitted a shadowed `length()` counterexample; strict base resolution retains the other 68. Plain-script and explicitly qualified exact-length guards are covered by regression tests and executable R witnesses.

Validation: new positive regression failed before the fix; full workspace tests, Clippy with warnings denied, formatting, and all 17 R oracle tests passed in this worktree's own target directory. Independent Astra review found the shadowing gap and verified its correction. Full Posit regeneration and both strict ecosystem checks pass with no committed report or message changes. Workspace tests and Clippy also pass after merging current main; the full R oracle is being repeated on that merged tree. CI and Pullfrog review remain required before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,9 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Avoid RY032 warnings when a parameter is only the lookup table for `%in%`,
+  or a base `length(x) == 1` guard protects a scalar predicate.
+
 - Report an explicit nesting-limit error before deeply nested syntax can
   overflow the parser stack. The limit is 128 tree-sitter syntax levels.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,9 +151,6 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


sims1253:
@pullfrog Please review the integration at 2195f99. Since reviewed 1656deb, the scalar release note moved beside the existing RY032 notes and approved #286 was merged from main. The scalar checker code and regression tests are unchanged. All current CI checks pass, as do local workspace, Clippy, formatting, full R oracle and both strict ecosystem checks. Please keep the incremental review to a single agent.

PR #330, opened 2026-09-07
Fix apply-family simplification and empty-input results

### LINKED ISSUE #321: Preserve list results when sapply or mapply cannot simplify
Two related false-positive cases remain in higher-order result simplification after the classed-input correction in #318. Both reproduce with candidate `92b3673` and pass their assertions in R 4.6.1.

Explicitly disabled simplification still receives an atomic result type:

```r
a <- sapply(1L, function(v) 1L, simplify = FALSE)
b <- mapply(function(v) 1L, 1L, SIMPLIFY = FALSE)
stopifnot(identical(a$field, NULL), identical(b$field, NULL))
```

ry emits two RY061 errors (`$ operator is invalid for atomic vectors of mode integer`). Both results are lists in R.

A plain input whose inferred length is unknown can also be empty:

```r
x <- integer(1L - 1L)
result <- sapply(x, function(v) 1L, USE.NAMES = FALSE)
stopifnot(identical(result$field, NULL))
```

ry emits RY061 here too. R returns an empty list. The scalar callback body cannot establish an atomic result when the input may be empty.

Follow-up acceptance:

- Respect matched `simplify` / `SIMPLIFY` controls, including named and reordered actuals; unknown controls must not establish a definite atomic result.
- Preserve the empty-list possibility when callback input cardinality is not proven nonzero, including mapply dots.
- Keep adjacent known-nonempty scalar simplification controls and required callback diagnostics.
- Add fresh R oracle witnesses and review any package/ledger changes by identity. Do not broaden the existing classed-input PR to absorb this follow-up.

sims1253:
Valid list results from `sapply`, `mapply`, and `tapply` could be inferred as atomic, causing false RY061 errors on `$` access. Respect matched simplification controls for proven base calls, including uncertainty from forwarded `...` and attached packages, and stop claiming atomic output when inputs may be empty. Custom stub contracts and callback diagnostics are preserved.

Regression tests and two R oracle fixtures cover named/reordered arguments, unknown and forwarded controls, empty inputs, custom stubs, and retained atomic results. Focused CLI checks also preserve ordinary data-frame callback errors and known-nonempty scalar results.

Workspace tests, clippy with warnings denied, formatting, and the full R oracle pass. No 500-package run was used; unknown-length inputs intentionally lose an unsupported atomic guarantee.

Closes #321.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,11 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Avoid claiming atomic `sapply()` and `mapply()` results when simplification
+  is disabled or uncertain, or when their inputs may be empty. Scalar
+  simplification requires a matched, enabled control and a provably nonempty
+  input.
+
 - Distinguish `@` slot extraction from `$`. Valid atomic `.Data` reads no
   longer trigger dollar-access errors. Keep slot results and replaced roots
   unknown, including mixed nested replacements; respect explicit accessors.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,10 +171,10 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
-- Avoid claiming atomic `sapply()` and `mapply()` results when simplification
-  is disabled or uncertain, or when their inputs may be empty. Scalar
-  simplification requires a matched, enabled control and a provably nonempty
-  input.
+- Avoid claiming atomic `sapply()`, `mapply()`, and `tapply()` results when
+  simplification is disabled or uncertain, controls are forwarded through
+  `...`, or inputs may be empty. Scalar simplification requires a matched,
+  enabled control and a provably nonempty input.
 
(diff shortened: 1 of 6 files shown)


sims1253:
The forwarded `...` finding is valid and is being fixed with focused regression coverage. We are also checking the related `tapply` control. The attached-package lookup correction is already in eda22d8, which landed after this review started.

The user explicitly ruled out another 500-package run. Validation will use focused positive/negative controls, workspace tests, and the R oracle; the conservative loss of precision for unknown-length inputs is disclosed and intentional. We will request review of the final follow-up commit.

sims1253:
@pullfrog Please review the follow-up commits through 7d7d49624a3ecb14838f66ad06dd27fd0e337b55. Both inline findings are addressed: forwarded `...` makes simplification uncertain, and the existing control handling now covers base `tapply`. The earlier eda22d8 commit also fixes attached-package uncertainty.

Workspace tests, clippy, formatting, full R oracle, and independent CLI positive/negative controls pass. Please use focused validation only; the user explicitly prohibited another 500-package run.

PR #391, opened 2026-09-08
feat(release): add linux-armhf and alpine (musl) release targets

sims1253:
Add release binaries and VS Code extensions for 32-bit ARM Linux and Alpine Linux on x64 and ARM64. Users of these platforms get a matching archive and platform-specific VSIX. The release matrix now covers nine targets.

| VS Code target | Rust target | Runtime check |
| --- | --- | --- |
| `linux-armhf` | `armv7-unknown-linux-gnueabihf` | Debian Bookworm ARMv7 under QEMU |
| `alpine-x64` | `x86_64-unknown-linux-musl` | Alpine 3.22, native |
| `alpine-arm64` | `aarch64-unknown-linux-musl` | Alpine 3.22 ARM64 under QEMU |

Update the changelog and release checklist for all nine targets. Zed continues to use its GNU Linux targets.

Validation:

- [Release dry run 34651525032](https://github.com/sims1253/ry/actions/runs/34651525032) passed all nine platform builds and executable checksums at `5356655`. `host` and `announce` were skipped. The later commit changes four remaining counts in the runbook.
- Downloaded and verified all nine archive checksums; independently recomputed all nine executable checksums and matched the uploaded sidecars.
- Built all nine Marketplace VSIXs locally. Verified each platform manifest and the bundled executable against the checked archive.
- Each new Linux binary prints its version and reports exactly RY061 for `x <- 1L; x$field` in its runtime environment. ARM checks used QEMU 6.2, not native ARM hardware.
- Installed the Alpine x64 VSIX in VS Code Server 1.137.0 on Alpine 3.22 and ran its installed binary.
- `dist generate --check`, release checksum tests, and extension type-check/build pass.

The integrated release candidate still needs its own dry run after the other sprint PRs are combined. No release or extension was published.


sims1253:
**0.9.1 release screening: Conditional; optional release scope.**

Green, latest automated review reports no blockers. Diff adds three core and VSIX targets. Release matrix is dynamic and checksum generation reads the configured targets; no generated-workflow defect was found. Add changelog/runbook updates and require a dry run of all nine archives, checksum sidecars, and platform smoke tests. C dependencies have not been qualified on the new paths by ordinary PR CI. Keep Zed's existing platform mapping.

Status refreshed against head `d7dd47b56eb568f606588fc80a63b7952192b4e4` on 2026-09-08. Screening is not a new full code review or release qualification.

**Admission gate:** update core release notes and every stale six-target core checklist; prefer referring to all configured targets. Qualify all nine archives and executable sidecars in a release dry run, then smoke-test armhf/Alpine binaries and compatible VSIX installation. PR CI and dist-plan generation do not establish successful cross-platform C-dependency builds. Keep this PR outside the stabilization milestone until this scope is admitted; existing-platform fixes need not wait.

Full work order and release gates: #392. [Milestone](https://github.com/sims1253/ry/milestone/1).

sims1253 (commented):
This PR grows the release matrix from six to nine targets: `armv7-unknown-linux-gnueabihf` (VS Code `linux-armhf`) and `x86_64`/`aarch64-unknown-linux-musl` (`alpine-x64`/`alpine-arm64`) are added to `dist-workspace.toml`, with matching entries in the hand-maintained `release-vscode.yml` matrix plus changelog and runbook count updates. `release.yml` is correctly untouched since its build matrix is computed at dispatch time by `dist plan`, and `dist generate --check` passes on this branch. I independently verified the plumbing: the dry-run musl artifacts (both x64 and arm64) are fully statically linked, `binary_checksums.py` is target-generic and its tests (which read the target list from the config) run in CI, the non-Windows download path in `release-vscode.yml` matches the `ry-cli-<target>/ry` archive layout, the Zed extension still pins gnu assets, and main has not gained these targets elsewhere (no conflicts or duplication).

## Findings

- [minor] `dist-workspace.toml:18` — The armhf artifact is dynamically linked against the ubuntu-22.04 builder's glibc: the dry-run binary (run 34651525032) needs `/lib/ld-linux-armhf.so.3` with symbol versions up to `GLIBC_2.34`, so Debian 11/Bullseye and Raspberry Pi OS Legacy 32-bit (glibc 2.31) will fail with a `GLIBC_2.34 not found` error. The Bookworm validation is consistent with this, but armhf users skew toward exactly those older 32-bit images (this is why ruff builds its armv7 target against manylinux2014/glibc 2.17). Since the shell installer will happily install an unrunnable binary on Bullseye, consider documenting the glibc ≥ 2.34 requirement in the runbook/README alongside the armhf archive.
- [nit] `.github/workflows/release-vscode.yml:81-86` — This matrix is now a hand-maintained copy of the `dist-workspace.toml` target list, and unlike `binary_checksums.py` (which reads the config directly), nothing fails if the two drift — a future tenth target would ship without its VSIXs and nothing would catch it. A small drift check comparing the matrix `dist-target`s against the config (e.g., alongside the existing `scripts/release` tests that already run in ci.yml) would lock the two together.
- [nit] `dist-workspace.toml:19-20` — The new targets implicitly ride cargo-dist 0.32's default runners (`ubuntu-22.04` for musl-x64 and the armhf cross, `ubuntu-22.04-arm` for musl-arm64 — confirmed in the dry-run job labels), which bypass the `[dist.github-custom-runners]` table the repo otherwise uses to pin macos/windows/xwin runners explicitly. Pinning the three new targets in that table would keep the runner set explicit and immune to cargo-dist default changes, in line with the repo's existing pinning discipline (and note that bumping off ubuntu-22.04 later will also raise the glibc floor from the first bullet).

## Overall

This is a well-validated change — the dry run, checksum sidecars, and Alpine runtime checks line up with what I could independently reproduce from the artifacts, and the musl/static-linking story is airtight. Merge-ready from my side; the only substantive ask is documenting the armhf glibc-2.34 baseline so 32-bit users on older distros aren't surprised.

Issue #381, opened 2026-09-08
RY070 must use R's function-mode call-head lookup (skip non-function bindings) — ~150 corpus FPs, sharpens to true no-function-exists errors


sims1253:
## Summary

RY070 models call heads with ordinary variable lookup, but R's function-mode lookup **skips non-function bindings in every frame** (`h <- 5; h()` → "could not find function \"h\"" — it never calls anything else, and never errors with "not a function"). Every RY070 of the form "local non-function value shadows an outward function" describes runtime behavior R cannot exhibit. ~99.3% of corpus RY070s are false positives; the mechanism families and the fix follow.

## Evidence

ry 0.9.0, 472-package corpus — 151 RY070 instances triaged at file:line, ~150 FP:

```r
h <- 5
h()    # R: Error in h() : could not find function "h"
       # ry: RY070 "h is double, not a function; cannot call it"  (wrong error shape)
```

Real-world instances of the impossible-shadow shape: `tweaks <- tweaks()` then later `tweaks(...)` (DBItest), `rank <- rank(x)` then `rank <= n` (forcats, round 1), `%<-%`/`<<-` placeholders (gitcreds, renv, lme4 `environment(f) <- rho`), NULL-R-CMD-check placeholders. In all of these, R resolves the call head by *function-mode* lookup: non-function bindings are invisible as call heads in every frame scanned — so if a function of that name exists anywhere outward, that is what runs.

The **1 true positive** found (lme4 `R/allFit.R:330` — unqualified `aes()` with only a NULL local placeholder and ggplot2 never imported) is the shape RY070 *should* exclusive target: no outward function exists at all. (Right diagnostic, wrong stated reason — like the ggplot2 `vjust` RY010 precedent.)

## Suggested direction

- For symbol call heads: apply function-mode lookup — skip non-function bindings per frame. RY070 on a symbol should fire **only when no function binding exists on any enclosing frame** (i.e., the call would be "could not find function"), which is also an error-severity claim that matches R's real error.
- Keep value-head checks (`obj$fn()`, `(expr)()`, literal heads) as-is — those genuinely error with "could not find function" or "not a function" in R.

This one semantic alignment eliminates most of the RY070 corpus mass (~150 FPs across ~51 packages) and sharpens the remaining diagnostics into "this call will definitely fail" — high-precision, error-severity.

## Context

[...]

PR #396, opened 2026-09-08
fix(checker): function-mode call-head lookup for bare symbols

### LINKED ISSUE #384: Typeshed dataset names (datasets::trees) outrank package-own exported functions (dials trees(), 16 FPs)
## Summary

Typeshed dataset names (e.g. `datasets::trees` the *data.frame*) outrank a package's own exported *functions* during symbol resolution inside that package — dials exports `trees()` (a tuning-parameter constructor) and every internal call `trees(...)` is flagged RY070 "trees is a list, not a function". Package-namespace bindings must outrank any typeshed name.

## Evidence

ry 0.9.0, 472-package corpus: **dials** — 16 RY070 instances on `trees()`, `tree`-adjacent constructors. The package's own NAMESPACE exports `trees`; `datasets::trees` is a built-in dataset the package neither attaches nor uses. ry's resolution order consults the datasets-stub name (via the open search path or base-attached datasets stubs) before the package's own exports.

Sister shape from the same resolution-order family (round-2 long-tail triage): `rownames(x) == seq_len(nrow(x))` style comparisons where base function names shadow columns are covered separately (#369); this issue is specifically **stub names beating package-own bindings**, the inverse of #346's "generic resolves to base".

## Suggested direction

Resolution order for a symbol inside package P must be: P's own definitions/exports (including NAMESPACE imports) → dependencies declared in DESCRIPTION → global/base. Typeshed *dataset* stubs should never participate in call-head resolution at all (datasets are data, not functions — see the function-mode lookup issue); at minimum they rank below everything package-local. Regression fixture: a package exporting `trees()` with internal `trees()` calls must produce zero diagnostics.

## Context

Found in the ry 0.9.0 corpus audit round 2 (artifacts `audit-20260908-ry090-r2/audit-results/ry070-ry061/`, mechanism M2).

sims1253:
## Summary

RY070 fired on bare symbol call heads from ordinary variable lookup. R uses function-mode lookup at call heads: non-function bindings are skipped in every frame, so a symbol call errors only when no function of the name is reachable anywhere. This gates both symbol-head RY070 sites (the lexical value stage and the bare typeshed-value stage) on a shared classifier:

- a dataset stub no longer beats a package-own exported function defined in another file (`trees()` in dials lost to the base `datasets::trees` stub);
- an open search path (library of an unstubbed package, unenumerable load) or an unknown data mask silences the diagnostic, decided before the retained callable-inventory carveout so uncertainty always wins;
- symbol heads report R's actual error, "could not find function `x` (bound here as `mode`)"; qualified `pkg::dataset()` and literal value heads keep their existing "attempt to apply non-function" wording and behavior.

The existing S7 callable_vars carveout is retained unchanged (approximation without execution-order information; refining frame-locality is deliberately out of scope). Known over-claims remain for a constructor bound in an outer function frame and for shadowed opaque outward bindings; both are pinned as documented gaps in tests.

## Tests

- 17 unit tests in `call_head_lookup` (outer function behind local value, dataset stubs vs project functions, uncertainty gates, qualified calls, masking, value heads, S7 carveout boundaries)
- 2 cross-file project tests (`cross_file_function_beats_dataset_stub` is the minimized #384 repro)
- R-executed oracle must-pass fixture `call_head_function_mode.R`
- corpus snapshot message updates; `docs/rules.md` RY070 row updated

Fixes #384. Partially addresses #381 (message wording, dataset stubs, uncertainty); the S7/frame-local shapes remain open there.

## Validation

- `cargo test --workspace`: 44/44 suites, 0 failures
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `cargo test -p ry-checker --test oracle -- --include-ignored` (R 4.6.1): 17 passed
- 0.9.0 dials-shape repro (NAMESPACE `export(trees)`, cross-file call) went from RY070 to zero diagnostics

Independent review: approved (exact-head review on the identical content, commit fade0659).


sims1253:
Published the reviewed follow-up: first-site uncertainty regression test (`172ff48`), native corpus reconciliation (`8ae4415`), and the separately audited label-only correction (`19c9978`). The corpus failures on the previous head are addressed by committed reports, messages, counts, hashes, and notes. Four RY070 identities disappear; the retained pkgdown templates finding remains emitted and is relabeled false positive based on pinned function-mode lookup evidence. No diagnostic is suppressed by the label correction. Standalone native checks and the local combined workspace preview pass. Fresh current-head CI/review and any main-integration reconciliation remain required.

Issue #342, opened 2026-09-07
Unknown-typed comparison operands collapse to logical<len=0> and fire RY001


sims1253:
## Summary

A comparison whose operand ry types as `unknown` is reported as `logical<len=0>`, firing RY001 for a condition that cannot be length zero at runtime. Unknown length should stay unknown, not collapse to 0.

## Evidence

ry 0.9.0 on ggplot2 @ 2d13fe6, `R/bin.R:101`:

```r
bin_breaks_bins <- function(x_range, bins = 30, center = NULL, boundary = NULL, closed = ...) {
  if (zero_range(x_range)) {
    width <- 0.1
  } else if (bins == 1) {     # RY001: condition is `logical<len=0>`
```

`ry dump-types` reports `bins -> unknown` in this scope, yet the diagnostic claims the condition is `logical<len=0>`. At runtime `bins` is always scalar here: the only caller (`compute_bins`, `R/bin.R:163-171`) validates it with `check_number_whole(bins, min = 1, allow_infinite = FALSE)` immediately before the call, and the parameter defaults to 30.

The len-0 shape appears to come from the interprocedural merge of the `bins` flow in `compute_bins` (parameter `NULL` | `allow_lambda(bins)` | function-applied `bins(x)`), collapsing to a zero-length numeric. Three isolated single-file repros of that caller flow stayed silent, so the trigger needs the whole-package fixpoint — but the invariant to fix is clear: a value whose type/length is unknown must not become "known length 0" as a side effect of joining call-site signatures.

## Related

Two adjacent refinements worth covering in the same work:
- `check_number_whole(bins, ...)`-style guard calls could refine the argument to a scalar for downstream analysis.
- The `else if` joins the flow after `zero_range(x_range)`, so the merge happens at a branch join; joins are where the collapse is observable.

## Context

Found in the ry 0.9.0 tidyverse corpus audit (artifacts: `audit-20260908-ry090/audit-results/ggplot2/`, including the dump-types output).

sims1253:
Corpus-scale confirmation (472-package run): the unknown→len0 collapse family is the largest RY001 mechanism — 83 instances (55%) of RY001 across the corpus, including a guard-then-flag shape where `if (is.null(val)) return()` immediately precedes the flagged `if (val > 1)` (gt ooxml2.R:612, datawizard row_count.R:107). pointblank alone has 37 via one unknown-returning helper (`get_threshold_type()`).

sims1253:
[...]

PR #397, opened 2026-09-08
fix(checker): a rejected mode test degrades default-parameter types

### LINKED ISSUE #343: RY061 flags $ on unknown/ggproto values in the else branch of is.character()
## Summary

RY061 reports ``$`` on "atomic vectors of mode `character`" for a value that ry itself types as `unknown`, because branch refinement around `if (is.character(x)) ... else ...` mishandles the else side for unknown-typed values.

## Evidence

ry 0.9.0 on ggplot2 @ 2d13fe6, `R/geom-boxplot.R:163-167`:

```r
if (is.character(position)) {
  if (varwidth == TRUE) position <- position_dodge2(preserve = "single")
} else {
  if (identical(position$preserve, "total") & varwidth == TRUE) {   # RY061 (x2, incl. $<- below)
    position$preserve <- "single"
  }
}
```

Both `$preserve` (line 164) and `` `$<-` `` (line 166) are flagged: "``$`` operator is invalid for atomic vectors of mode `character`". Runtime-verified false positive: in the else branch `position` is a Position ggproto object (an environment); constructing `geom_boxplot(position = position_dodge2(preserve = "total"), varwidth = TRUE)` executes the read and emits ggplot2's intended "Can't preserve total widths" warning.

`ry dump-types` reports `position -> unknown` at function scope, so the "mode character" claim is manufactured inside the branch analysis. In the else branch of `is.character(position)`, the value is *not* character — for an unknown-typed input the refinement should stay unknown (or at worst "not character"), never character. Isolated repros of the nested-reassignment shape stayed silent, so the exact merge path likely involves the conditional reassignment (`position <- position_dodge2(...)`) inside the then-branch plus the join afterwards.

## Suggested direction

Audit the negated-predicate refinement lattice (`!is.character(x)`) for unknown inputs: negation of a positive type test must never *introduce* the tested mode. A regression test shaped like the snippet above with an unknown-typed parameter would pin it.

## Context

Found in the ry 0.9.0 tidyverse corpus audit (artifacts: `audit-20260908-ry090/audit-results/ggplot2/`).

sims1253:
Fixes #343
Milestone: 0.9.1 — stabilization

## Problem

A defaulted parameter's recorded type describes only the omitted-argument call shape (`Checker::diagnostic_parameter_type`). ggplot2's `geom_boxplot` (`R/geom-boxplot.R:163-167`) gains `character` for `position` whenever a testthat call site omits it, and the else branch of `if (is.character(position))` kept that mode, so the `$preserve` read and write on the Position ggproto object fired RY061 twice:

    RY061: $ operator is invalid for atomic vectors of mode `character`

Reproduced on ggplot2 @ 2d13fe6 with the v0.9.0 baseline; runtime-verified false positive (the else branch holds a Position ggproto environment; ggplot2's own "Can't preserve total widths" warning path executes the read).

## Change

In `apply_narrowing_branch` (`crates/ry-checker/src/infer/narrow.rs`), the `Positive`/`Else` arm (previously NULL-complement only) now also degrades a default-parameter binding to unknown in that branch when a mode predicate is rejected and the recorded type is exactly the tested mode family — single mode equal, or a union whose every member mode is among the target's member modes (`is.numeric`'s integer|double family included).

Guarded to mode predicates only: the target must carry a concrete or union mode and no explicit class. `Mode::Opaque` targets are class predicates (`inherits`, `is.environment`, `is.<class>`), and a concrete mode target can also carry an explicit class (`is.data.frame` is list + "data.frame"; stub-declared predicates can express the same shape) — a false path there rejects the class claim, not the mode, so `has_known_class()` targets never degrade.

## Boundaries left untouched (deliberate)

- Class predicates (`inherits`, `is.data.frame`, stub predicates with class) never degrade; pinned by `class_predicate_else_branch_keeps_its_diagnostic` and `class_carrying_predicate_targets_do_not_degrade_the_rejected_branch` (red-first: fails without the class exclusion).
- Local literals keep their type: their rejected branch is genuinely unreachable (`local_literal_else_branch_keeps_its_unreachable_diagnostic`).
- Mixed unions keep any surviving member (`union[character, integer]` rejected against `is.character` still justifies `$` diagnostics); pinned by the helper matrix `exactly_tested_family_covers_groups_but_not_mixed_unions`.
- The NULL complement path is unchanged.
- Rebinding clears the default-parameter marker before this can apply (`reassigned_formal_keeps_its_else_branch_diagnostic`).
- The continuation after the join still keeps the default-derived shape (`continuation_after_branch_keeps_default_shape_diagnostic`); join-time default treatment is out of scope.

## Tests

Red-first on baseline: unit reds for the character default and the `is.numeric` family; package reds on the exact ggplot2 `geom_boxplot` shape (`tests/default_parameter_branches.rs`) for an omitting testthat call site and for mixed call shapes. All controls listed above stayed green throughout.

## Evidence

- ggplot2 @ 2d13fe6 (346 files): exactly the two `geom-boxplot.R` RY061 disappear; every other finding is unchanged (issues #342/#373 handled separately).
- Local gates run on the published head bade704f16e3f7f24b546ae81c72d16d3c292a00a: `cargo test -p ry-checker --lib` (723 passed), `cargo test -p ry-checker --test default_parameter_branches` (3 passed), `cargo test -p ry-checker --test oracle` (16 passed, 1 ignored), `cargo clippy -p ry-checker --all-targets -- -D warnings` (clean), `cargo fmt --all -- --check` (clean). No other gates claimed.


sims1253:
Published corpus repair `2ee3ca3`. Native comparison removes exactly the two reviewed ggplot2 RY061 false positives at R/geom-boxplot.R:164:19 and :166:7 in both corpora, with no additions or true-positive changes. Reports, messages, counts, notes, producing revision, hashes, and README are reconciled. Both committed-tree native checks pass. Checker code is unchanged from the previously reviewed head. Fresh current-head CI/review and final main integration remain required.

Issue #354, opened 2026-09-07
Diagnose NA-valued if/while conditions and NA propagation into conditions (RY001 family)


sims1253:
## Summary

ry does not diagnose NA-valued `if`/`while` conditions — not even a literal `if (NA)` — and four shipped defects in the corpus audit crash precisely there ("missing value where TRUE/FALSE needed"). An RY001-family extension covering NA (and NA-producing comparisons/coercions feeding conditions) is warranted.

## Evidence

Probe: `ry check` on a file containing `if (NA) 1` emits nothing (RY001 covers non-length-1 logicals but not NA).

Runtime-verified instances in shipped code (ry 0.9.0 silent on all):

1. **tidyr @ 26f83e8, `R/seq.R:24-29`** — `full_seq(c(1, NA, 3), 1)` dies: the level check `any(((x - rng[1]) %% period > tol) & ...)` propagates NA into `if` although `range(x, na.rm = TRUE)` two lines up anticipates NAs. No NA-input test exists upstream.
2. **lubridate @ 9aa03c4, `R/intervals.r:525`** — `setdiff(as.interval(NA), i1)` aborts with "argument is not interpretable as logical": `if (sum(makes2))` where the summands can be NA; sibling `intersect()`/`union()` guard gracefully.
3. **readxl @ 47f8aea, `R/read_excel.R:301`** — `if (sheet < 1)` with `sheet = NA_real_` (passes `is.numeric`) produces the raw runtime error instead of the intended clean `stop()`.
4. **haven @ f067fb2, `R/haven-stata.R:190,244`** — `as.integer()` overflow turns label values into NA; `x_finite == as.integer(x_finite)` yields NA; `if (any(bad_labels))` dies before the intended `cli_abort`.

Common shapes worth covering, in rough order of value:
- literal `NA` reaching a condition;
- `if (any(x))`/`if (sum(x))` where `x`'s element type is NA-able and unguarded (`==`, `>`, arithmetic on possibly-NA);
- comparisons whose operands pass through NA-introducing coercions (`as.integer` of out-of-range doubles is the concrete overflow case).

RY034 already covers literal `== NA`; this is its dataflow generalization.

## Context

Found in the ry 0.9.0 tidyverse corpus audit (artifacts `audit-20260908-ry090/audit-results/{tidyr,lubridate,readxl,haven}/`, each with a deterministic repro).

sims1253:
**0.9.1 release screening: Later; narrow first slice.**

[...]

PR #398, opened 2026-09-08
fix(conditions): accept R-coercible scalar if/while conditions

### LINKED ISSUE #373: if/while conditions accept coercible character/numeric (Sys.getenv flags) — RY001 message and semantics wrong
## Summary

`if (Sys.getenv("FLAG"))` is a widespread, working R idiom: R coerces the character `"TRUE"`/`"FALSE"` (and numeric) condition values at `if`/`while`. ry reports "condition is `character<len=1>`, expected length-1 logical" — factually wrong about R semantics — and the corpus shows ~14 sites relying on it. 10 RY001 instances in the corpus plus 14 RY033 sibling sites.

## Evidence

ry 0.9.0, 472-package corpus: 10 RY001 instances of `if (<character condition>)`, predominantly `Sys.getenv(...)` flags; Rscript-verified that R accepts `if ("TRUE") 1` (and errors helpfully for other strings). Sibling RY033 sites (`Sys.getenv("RSTUDIO") == 1`) add ~14 more instances of the same env-flag idiom family.

The message's claim "expected length-1 logical" is not what R requires: R requires a length-1 logical *or* something coercible to it (character "T"/"TRUE"/"true"/numeric). The diagnostic is both a false positive and semantically incorrect as stated.

## Suggested direction

Two parts:
1. Model R's condition coercion: `character<len=1>` and `numeric<len=1>` conditions are legal when their values are in the coercible set; since values are usually unknowable statically, the right move is silence (or at most an info-level note for conditions that are *provably* non-coercible strings).
2. Fix the message wording wherever condition-type diagnostics fire — "expected length-1 logical (or coercible)" — so remaining true positives don't misstate R.

## Context

Found in the ry 0.9.0 corpus audit round 2 (artifacts `audit-20260908-ry090-r2/audit-results/ry032-ry001/`, mechanism M5; Rscript coercion verification included).

sims1253:
## Summary

R coerces `if`/`while` conditions beyond length-1 logicals: numeric and raw scalars coerce numerically, complex scalars coerce by OR over their real and imaginary components, and a character scalar coerces only when its text is one of `T`/`TRUE`/`true`/`True`/`F`/`FALSE`/`false`/`False`. ry reported every scalar character condition as invalid — most visibly the widespread `if (Sys.getenv("FLAG"))` idiom — and the RY001 message ("expected length-1 logical") misstated R. This closes #373.

## Behavior

- Scalar (length-1 or unknown-length) character, complex, and raw conditions are no longer RY001. Zero-length and known length > 1 conditions stay RY001 regardless of mode; list, NULL, and function conditions stay RY001.
- A direct string literal is still decidable: the parser decodes escapes (hex/octal/unicode) and raw strings, so any literal outside the eight accepted spellings (`"x"`, `"tRue"`, `"1"`, `""`, `"NA"`, ...) keeps RY001.
- Mixed unions of a numeric member and any coercible scalar member (character/complex/raw) stay silent; a proven-invalid member still dominates. Pure numeric unions keep RY003; the existing logical/opaque union silence is unchanged.
- RY001 wording: "expected length-1 logical (or a value R coerces to logical)"; rules registry summary and docs/rules.md updated.

## Boundaries (not claimed)

- `Sys.getenv("FLAG")` with the variable unset returns `""` and errors at runtime in R; silence for unknown-value scalar strings is the uncertainty policy, not a claim that R always accepts them.
- NA stays out of scope (#354): `as.logical("NA")` returns NA and using it as a condition errors; the `if`/`while` coercion path rejects the `"NA"` string earlier, in the same family as other non-literal text. `NA_complex_` stays silent as an untracked-value uncertainty (R errors on it at runtime; the oracle pins that fact without treating it as valid).
- RY033 sibling sites (`Sys.getenv("RSTUDIO") == 1`) are untouched.

## Tests and evidence (R 4.6.1 oracle)

- New corpus fixtures: `ok_coercible_condition.R` (0 diagnostics; previously 12x RY001) and `warn_noncoercible_condition_shapes.R` (10 retained RY001 rows: non-literal strings, zero-length, length-2 character/complex, list, NULL).
- New oracle fixture `condition_coercion.R`: pins the eight accepted spellings (incl. hex/octal/raw-string forms), raw TRUE/FALSE branches, complex OR-truthiness (`if (0+0i)` else-branch, `if (0+1i)`/`if (2+3i)` then-branch), and the runtime errors for `"tRue"`, `"NA"`, `NA_complex_`, zero-length and length-2 complex.
- Unit tests: silent rows (literals, escapes, Sys.getenv if/while/binding, raw, complex constructors, NA_complex_ boundary), retained-invalid battery, union contract (mixed silence incl. red-first numeric|raw row; invalid dominance; logical/opaque silence; pure numeric RY003), and mode-provenance assertions that complex/raw paths are real `Mode::Complex`/`Mode::Raw`, not opaque fallbacks.
- Two ry-lsp protocol probes that pinned the old character-condition contract via stub character returns now use list returns so they keep proving stub isolation and config degradation.

## Gates

On the exact head: `cargo test --workspace` (54 suites, 1389 passed / 0 failed), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `cargo test -p ry-checker --test oracle -- --include-ignored` (17/0) all clean.

## Package delta vs published v0.9.0

- #373 repro package: 4 -> 2 warnings (both `Sys.getenv` FPs removed; the two proven-invalid literals retained).
- ggplot2 @ 2d13fe6 (346 files): identical output except the RY001 message wording on `R/bin.R:101` (a #342 zero-length shape, untouched); totals unchanged.
- googlesheets4 @ 55cd9fd (104 files): byte-identical.

Closes #373.

## Summary by CodeRabbit

* **Bug Fixes**
  * Condition analysis now recognizes additional scalar values that R can coerce to logical, reducing false-positive RY001 warnings.
  * RY001 remains reported for provably invalid conditions, including unsupported strings, invalid lengths, lists, and NULL.
  * Updated RY001 messaging to reflect accepted coercible conditions.

* **Documentation**
  * Updated rule documentation and diagnostic reports to reflect revised condition behavior.

* **Tests**
  * Expanded coverage for character, raw, complex, numeric, union, and invalid condition cases.


sims1253:
Published corpus repair `ece9ed2` and stale-comment correction `9452a21`. Both LSP test comments now describe the actual list-returning controls. Native comparison removes exactly four audited renv RY001 sites; their former TP labels assumed scalar strings could not be conditions. Pinned R expressions accept the default TRUE/FALSE values and reject garbage; unknown external input is not a statically proven error. All other finding identities are retained; 25 RY001 messages gain only the coercion clarification. Reports, messages, counts, notes, and hashes are reconciled, and both committed-tree native checks pass. This does not claim arbitrary environment values are valid. Fresh current-head CI/review and final main integration remain required.

sims1253:
Reconciled with accepted main (#397 and #400) using normal merges. The corpus reconciliation preserves the four audited renv removals and all retained labels.

Addressed the new review findings:
- Known lengths above one are invalid before mode classification. Numeric `if` and numeric/logical `while` conditions now emit RY001 instead of RY003 or silence. Logical `if` retains its established RY002 diagnostic, without duplicate family diagnostics.
- A six-case regression failed before the repair and passes after it. The R oracle confirms that all six conditions fail at runtime.
- The union-condition test now asserts the inferred union/scalar shape.
- Source and oracle comments use R's exact error wording.

Workspace tests, clippy, formatting, and the runtime oracle pass. Both corpora were regenerated with the immutable binary from `ccdedb07d2103b92b3a57554cf464fdd514d5493`. The length repair produces no report-byte, identity, or label changes from the integrated corpus. Producer metadata now records that source.

Final counts: Posit 443 (37 TP / 406 FP); tidyverse 103 (10 TP / 41 FP / 52 unowned). Fresh current-head review and CI are required before merge.

sims1253:
Disposition of the outside-diff union comment: retain the existing uncertainty policy, with explicit coverage for the reported cases.

I tested the proposed scan-all-members change. It made the logical/list regression emit RY001, but failed the existing `tests::scope_resolution::condition_union_with_a_valid_logical_member_is_silent` test. That test requires `logical(0) | TRUE` to remain silent: “a possibly-valid condition must not be reported as definitely invalid.” It predates this PR (already present in the #124 test-module split), and accepted main has the same early return.

The behavior change is withdrawn; the existing test is unchanged. New tests pin logical/list unions in both branch orders for both `if` and `while`, plus logical/numeric silence. The R oracle confirms both outcomes: the logical branch passes and the list branch fails. Silence is an uncertainty boundary, not a claim that the list branch is valid. This PR retains the established logical/opaque-alternative policy; changing it requires an explicit policy decision rather than weakening a retention test.

Workspace tests, clippy, formatting, and the runtime oracle pass. This follow-up changes only tests and the oracle, not executable checker source or corpus artifacts. The previously validated `ccdedb07` binary and report provenance remain applicable. Fresh current-head review and CI are still required.

PR #400, opened 2026-09-09
fix(checker): bound forwarded-default invalidation to root calls

### LINKED ISSUE #342: Unknown-typed comparison operands collapse to logical<len=0> and fire RY001

sims1253:
## Summary

Fixes #342 with a bounded forwarding change. A possible rebinding in a statement before a bare root-level forwarding call invalidates its default-derived fact. Definitely later statements do not invalidate the fact.

Only an exact root-call node/span receives an ordering anchor. Wrapped, nested, loop-contained and branch-contained calls retain their pre-fix behavior. This is not a general control-flow or promise-forcing analysis. Same-statement bindings and deferred forcing remain outside this bounded rule.

## Regression evidence

- Against the immutable published v0.9.0 binary: the rebinding false positive disappears; straight-line post-call, wrapped-call, loop-call, branch and ordinary true-positive controls remain.
- Root-node/span anchoring has a dedicated collector unit test. Cross-file package fixtures exercise both false-positive removal and retained controls.
- Native pinned comparisons remove exactly ggplot2 `RY001` at `R/bin.R:101:14` in both corpora. All other identities are retained, including all 28 remaining Posit RY001 sites on the original standalone comparison.
- Reports, message ledger, classification counts, audit notes, README and source hashes are reconciled on accepted main, including #395/#396. Posit: 449 findings, 41 TP / 408 FP. Tidyverse: 105 findings, 10 TP / 43 FP / 52 unowned.

## Validation

- Exact integrated-head `cargo test --workspace`: pass.
- Standalone bounded code: independent review, clippy, fmt and R oracle controls passed. The combined local preview also passes workspace, clippy, oracle, 1,000 LSP seeds, fuzz and tracked-only archive build; this is supporting evidence, not a substitute for this PR's current-head CI.
- Final integrated pinned tidyverse check: pass. Final integrated Posit full check: pass (447 distinct readable identities over 449 ledger rows; the two pre-existing duplicate identities are preserved).

History uses normal merges; no source or ledger suppression is introduced to bypass CI. Fresh pullfrog review and required CI remain merge gates. No release, tag or Pages change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RY001 diagnostic accuracy when a forwarded parameter is reassigned before a direct forwarding call.
  * Prevented false-positive diagnostics for assignments, loop variables, and related rebinding patterns.
  * Preserved existing diagnostics for wrapped, nested, or post-call rebinding scenarios.

* **Documentation**
  * Updated ecosystem diagnostic summaries and corpus records to remove the corrected false positive and refresh counts.


sims1253:
Addressed in `2a36492242676100f9ad246fe5740b9e12f32ed2`:

- Normalize both the source name and assignment/loop/expression targets through `semantic_argument_name`.
- Added a helper regression covering both bare/backticked directions, all three assignment arms, unrelated names, and the before-call boundary. Added end-to-end backticked assignment, loop-variable, and condition-assignment controls.
- Both new tests failed on the prior head with the reported RY001 and pass after the fix.
- Corrected the loop test comment: it retains a known bounded-scope false positive, not a true-positive control. The wrapped-call retention control remains distinct.
- Documented dynamic `assign()` and replacement targets as remaining limitations.

Validation: workspace tests, clippy, fmt, and both native pinned corpus checks pass. Reports and ledger identities are unchanged. Fresh current-head review/CI remain required.

Issue #366, opened 2026-09-08
NAMESPACE wholesale import(pkg) unresolved — every imported symbol reported unbound (RY010, 322+ corpus FPs)


sims1253:
## Summary

NAMESPACE wholesale imports (`import(pkg)` — import *all* of a package's exports) are not resolved to their source bindings, so every symbol coming through such an import is reported as an unbound variable (RY010). This single mechanism accounts for 322+ corpus instances at essentially 100% false-positive rate and puts two packages at the top of the diagnostic-density leaderboard.

## Evidence

ry 0.9.0, 472-package corpus run (2026-09-08):

- **param6** — 133 diagnostics, all RY010 on `Set`/`Reals`/`Interval`/`PosReals`/... — every diagnostic in the package. Root cause verified at source: `NAMESPACE` contains `import(set6)` (`R/support_dictionary.R:13` uses the symbols unqualified), and set6 exports them. Density: 886.7 diagnostics per 100 files — rank 1 of the corpus, entirely this mechanism.
- **distr6** — rank 4 (191 diagnostics), same `Set`/`Interval`/`Reals` family via the same import style.
- Family total: ~322+ instances of the top RY010 name clusters (`Set`, `Interval`, `Reals`, `PosReals`, `Naturals`, `CondEqual` and siblings) across the set6/distr6/param6 packages, all traced to wholesale imports.

`importFrom(pkg, sym)` entries are resolved correctly (round 1 confirmed bare `map_lgl` via importFrom triggers rules — #365 covers its own gap); it is specifically the un-prefixed `import(pkg)` catch-all form that loses the bindings.

## Suggested direction

When NAMESPACE declares `import(<pkg>)`, every export of `<pkg>` should enter the package-scope symbol table with unknown type (silencing RY010 and letting other rules treat them as opaque). Export lists are statically derivable from the dependency's NAMESPACE (or its typeshed). A regression fixture: a package with `import(sets)`-style NAMESPACE and unqualified uses of 3+ exported symbols must produce zero RY010s.

## Context

Found in the ry 0.9.0 corpus audit round 2 (472 packages; artifacts `audit-20260908-ry090-r2/audit-results/profile/`, cluster detail in `ry010/`).

sims1253:
**0.9.1 release screening: Investigate → core if bounded.**

Wholesale imports already modeled; reproduce export-inventory availability and namespace precedence gaps.

**Evidence and scope correction:**

[...]

Issue #40, opened 2026-08-06
r-typeshed schema fields for hardcoded checker lists (pattern audit followup)


sims1253:
## Context

The pattern audit (`docs/plans/31-pattern-audit.md`) found 15 hardcoded function-name lists in ry-checker where each member carries an implicit semantic claim about R behavior. `SCALAR_REDUCTIONS` was migrated to use the stubs' existing `return.length: "1"` data. Several other lists can't be migrated because the stubs schema has no field to express the claim.

### `predicate_target` (12 `is.*` functions)
- **Claim:** each predicate tests for a specific `RType`, so an `if` condition using one narrows the variable's type.
- **Stubs have:** return type (`logical`, length 1).
- **Missing:** the *target* type the predicate tests for (e.g. `is.numeric` → narrows to `double|integer`).
- **Proposed field:** `"narrows_to": { "mode": "double_or_int" }` on `FunctionSig`.

### `FFI_PRIMITIVES` (5 functions)
- **Claim:** first argument is a native entry-point symbol, not a value reference.
- **Stubs have:** params and return types.
- **Missing:** the first-argument-is-symbol convention.
- **Proposed field:** `"native_symbol_arg": 0` on `FunctionSig` (zero-based arg index).

### `s3_group_generic` (34 Math/Summary functions)
- **Claim:** calling on a classed value dispatches to `Math.foo` / `Summary.foo`.
- **Stubs have:** group generic names in `globals.s3_generics`.
- **Missing:** the membership mapping ("`abs` dispatches to `Math`").
- **Proposed field:** `"s3_group": "Math"` on `FunctionSig`.

### `assign_replacement_target` (7 functions)
- **Claim:** replacement-function generics (`names(x) <- v`) rebind the first argument.
- **Stubs have:** function stubs exist.
- **Missing:** the replacement-function property.
- **Proposed field:** `"is_replacement_function": true` on `FunctionSig`.

### `DEFUSING_CALLS` (7 remaining members without stub `eval`)
- `expression`, `quote`, `substitute`, `bquote`, `alist` in base
- `expr`, `exprs`, `quo` in rlang
- Add `"eval": { "...": "captures_promise" }` or the appropriate per-parameter mode.

### `expression_has_list_origin` (3 names: `list`, `lapply`, `Map`)
- Already declare `return.mode: "list"` in stubs; checker needs threading typeshed into the standalone `expression_has_list_origin` fn.

[...]

PR #441, opened 2026-09-11
fix(typeshed): resolve set6 and dictionar6 imports

sims1253:
Packages with `import(set6)` or `import(dictionar6)` report valid exports as unbound when those dependencies are not installed. Vendor the complete export inventories from [r-typeshed #41](https://github.com/sims1253/r-typeshed/pull/41), register both packages, and test the CLI with installed-library lookup disabled.

```text
import(set6) → bundled exports → Set, Reals, Interval resolve
importFrom(set6, Set)          → only Set resolves
                              → Reaals still reports RY010
```

The stubs keep R6 generators opaque and function formals inference-only. The upstream runtime tests compare all exports and formals against set6 0.2.4 and dictionar6 0.1.3. This PR records the clean upstream commit in `vendor/SOURCE`; merge the typeshed PR first.

On the pinned sources from #366, param6 0.2.4 goes from 133 diagnostics to zero and distr6 1.6.9 from 191 to 10. All 314 removed findings are RY010s for verified exports. No diagnostic is added, and all remaining findings match the baseline. Source hashes, reproduction commands, and limits are in `docs/corpus/set6-inventory.md`. Other missing dependency inventories remain outside this change.

A wider comparison used the same baseline binary with only the two new stub directories as overrides across 472 local package checkouts. Diagnostics fell from 3,106 to 2,784: 322 RY010s removed, none added. Only param6 and distr6 changed; that panel uses distr6 1.8.4, which accounts for the larger reduction. Every removed name is a verified export. The bundled binary produces identical results for both reported packages and for the changed corpus packages.

Validation:

- The new CLI regression fails on `92c822a` and passes with the vendored stubs. It covers wholesale imports, selective imports, absent imports, and a misspelling.
- `cargo test --locked --workspace`, Clippy with warnings denied, formatting, and the full R oracle (`--include-ignored`) pass.
- Typeshed validation passes for all 38 stub files. The sync harness and release-checksum unit tests pass.
- The tracked tidyverse corpus matches its committed reports. All GitHub CI jobs pass, including the Posit corpus and ledger regression checks. The extra local full-tier Posit run was stopped after CI passed; it is not counted as a completed check.
- Pullfrog reviewed this commit and the upstream typeshed PR and found no new issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Bundle set6 and dictionar6 export inventories so wholesale imports resolve
+  their R6 objects and functions without installed copies of those packages
+  (#366). Keep unrelated unknown names reportable.
 - Discard forwarded-default facts after writes before wrapped or nested calls,
   including replacement assignments, loop variables, and literal `assign()`
   targets. Match backticked parameter and argument names consistently
--- a/crates/ry-cli/tests/config_e2e.rs
+++ b/crates/ry-cli/tests/config_e2e.rs
@@ -962,6 +962,64 @@ fn package_namespace_whole_import_binds_installed_exports() {
 }
 
 #[test]
+fn bundled_set6_imports_bind_exports_without_hiding_misspellings() {
+    for (namespace, unbound) in [
+        ("import(set6)\nimport(dictionar6)\n", vec!["Reaals"]),
+        (
+            "importFrom(set6, Set)\nimportFrom(dictionar6, Dictionary)\n",
+            vec!["Reals", "Interval", "as.Set", "dct", "Reaals"],
+        ),
+        (
+            "",
+            vec![
+                "Set",
+                "Reals",
+                "Interval",
+                "Dictionary",
+                "as.Set",
+                "dct",
+                "Reaals",
+            ],
+        ),
+    ] {
+        let tmp = tempfile::tempdir().unwrap();
+        fs::create_dir(tmp.path().join("R")).unwrap();
+        fs::write(
+            tmp.path().join("DESCRIPTION"),
+            "Package: inventoryfixture\nVersion: 0.0.1\nImports: set6, dictionar6\n",
+        )
+        .unwrap();
+        fs::write(tmp.path().join("NAMESPACE"), namespace).unwrap();
+        fs::write(
+            tmp.path().join("R/use.R"),
+            "values <- list(Set, Reals, Interval, Dictionary, as.Set, dct, Reaals)\n",
+        )
+        .unwrap();
+
+        let output = Command::new(env!("CARGO_BIN_EXE_ry"))
+            .args(["check", "--output-format", "json"])
+            .arg(tmp.path())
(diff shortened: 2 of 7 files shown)


sims1253 (commented):
This PR resolves the false RY010s from #366 by bundling export inventories for set6 and dictionar6: the two stub JSONs are vendored from r-typeshed#41 (`vendor/SOURCE` bumped to `cadfe7a`), both packages are registered in `packages.txt`, a corpus doc records sources/checks, and a hermetic CLI regression test (run with `RY_NO_INSTALLED_LIBRARIES=1`) covers wholesale `import()`, selective `importFrom()`, absent imports, and a misspelling. No checker code changes — registration goes through the existing generic path, which I verified: `crates/ry-typeshed/build.rs` generates `PACKAGE_SPECS` from `packages.txt` and hard-asserts every registered name has a vendored stub, and workspace resolution unions each attached/whole-imported package's stub function+dataset keys with installed-library exports.

## Findings

- [nit] `crates/ry-cli/tests/config_e2e.rs:968-971` — the `importFrom` sub-case doesn't actually depend on the bundled stubs: selective-import names are bound unconditionally from the NAMESPACE directive (`crates/ry-workspace/src/packages.rs`, `importFrom` arm) and extended into `file_bindings` at `crates/ry-workspace/src/lib.rs:234` without consulting any inventory, so that case passes on the base commit too. Fine as behavior pinning, but only the `import(set6)` case is the regression this PR's data fixes (which the PR body's `92c822a` failure note confirms).
- [nit] Same area — now that an export inventory is bundled, `importFrom(set6, Reaals)` still binds the misspelled name silently; only the bare use is flagged. Validating selective imports against known inventories (when one exists) would be a cheap follow-up hardening.
- [question] Both stubs declare `"version": "0.0.1"` (matching the R6/vctrs opaque-inventory convention) while documenting inventories from set6 0.2.4 / dictionar6 0.1.3, and resolution unions installed exports with the bundled keys (`crates/ry-workspace/src/lib.rs:346-356`), so bundled names bind even when a different version is installed and vice versa. `docs/corpus/set6-inventory.md` pins the audited versions and the 472-package panel; just confirming the union semantics are the intended long-term tradeoff.
- [nit] No `ry-typeshed` unit test asserting sentinel exports for the two new packages (cf. `load_package_vctrs_has_hermetic_imports`). `every_known_package_loads` validates parse/registration and the e2e covers behavior, so this is optional and matches convention only loosely.

Cross-repo consistency verified: all 38 vendored blobs on this branch are byte-identical (matching git blob SHAs) to `stubs/` at r-typeshed commit `cadfe7a` (PR #41's head), the `SOURCE` provenance matches, and the doc's `--typeshed stubs/set6` replay paths match the partner repo layout and `load_stub_dir`'s nested-layout support. `cadfe7a` will remain reachable after #41 merges given r-typeshed's merge-commit style (the previous `SOURCE` commit `a682775` is exactly PR #40's merge commit).

## Overall

This is a clean, data-only fix that uses the generic registration path exactly as designed, with hermetic test coverage and honest documentation of its limits; I found no correctness issues in the resolution path. The findings above are minor polish items, none blocking.

PR #442, opened 2026-09-11
fix(checker): align condition diagnostic severity

### LINKED ISSUE #415: RY001/RY002 family consistency: rule-table severity vs emitted severity, and RY100 gating of RY002
$3

sims1253:
The checker API returned RY001 as an error while the rule table and CLI exposed it as a warning. Return a warning at the source so API consumers receive the documented severity.

Pin the existing RY002 behavior when a condition also contains RY100: the length warning remains because the multi-value condition still errors in R. Update the two RY001 oracle fixtures to assert their runtime errors and expect the documented warning. The corpus snapshot changes only the severity of 13 existing RY001 findings.

Closes #415.

Validation: workspace tests, Clippy with warnings denied, formatting, and the full R oracle pass. CLI diagnostic output keeps its existing severity.

## Summary by CodeRabbit

* **Bug Fixes**
  * RY001 condition diagnostics are now consistently reported as warnings across the checker API, rule table, and command-line output.
  * RY002 length warnings are retained when the same condition also triggers RY100.
  * Improved handling of forwarded and reassigned values in condition checks, with clearer runtime error reporting in affected scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Return RY001 warnings from the checker API, matching the rule table and CLI.
+  Keep RY002 length warnings when the condition also contains RY100 (#415).
+
 - Discard forwarded-default facts after writes before wrapped or nested calls,
   including replacement assignments, loop variables, and literal `assign()`
   targets. Match backticked parameter and argument names consistently
--- a/crates/ry-checker/src/infer/mod.rs
+++ b/crates/ry-checker/src/infer/mod.rs
@@ -631,7 +631,7 @@ impl Checker {
             && !has_ry100
         {
             self.emit(
-                Severity::Error,
+                Severity::Warning,
                 span_of(cond),
                 "RY001",
                 format!(
--- a/crates/ry-checker/src/tests/type_inference.rs
+++ b/crates/ry-checker/src/tests/type_inference.rs
@@ -1956,13 +1956,14 @@ fn quoted_symbol_existence_does_not_invent_distinct_bindings() {
 fn condition_severities_match_the_rule_registry() {
     for source in [
         "if (NULL) 1L",
+        "if (1L) 1L",
         "if (c(TRUE, FALSE)) 1L",
         "while (list(TRUE)) break",
     ] {
         let diagnostics = check(source);
         let condition = diagnostics
             .iter()
-            .find(|d| matches!(d.code, "RY001" | "RY002"))
+            .find(|d| matches!(d.code, "RY001" | "RY002" | "RY003"))
             .unwrap();
         assert_eq!(
             condition.severity,
--- a/crates/ry-checker/src/tests/type_inference.rs
+++ b/crates/ry-checker/src/tests/type_inference.rs
@@ -1951,3 +1951,36 @@ fn quoted_symbol_existence_does_not_invent_distinct_bindings() {
         );
     }
 }
+
+#[test]
+fn condition_severities_match_the_rule_registry() {
+    for source in [
+        "if (NULL) 1L",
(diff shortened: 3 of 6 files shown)


sims1253 (commented):
This PR fixes the severity mismatch for the condition-diagnostic family: the checker emitted RY001 (invalid-condition) as `Severity::Error` from the raw API while the rule registry, `docs/rules.md`, and the CLI (which re-severities through `apply_filter_to_diagnostics`) all exposed it as a warning. The single-line change at the emit site aligns the API with the documented severity, pins the RY002 behavior when a condition also contains RY100 with a new test, converts the two RY001 oracle fixtures from `must-flag` to `must-warn RY001` (making the R-side runtime error explicit via `tryCatch` assertions), and updates the 13 affected corpus-snapshot severities.

## Findings

- [question] `crates/ry-checker/src/infer/mod.rs:634` — the unified warning also downgrades RY001 subsets that unconditionally error in R (`NULL` → "argument is of length zero", list/complex conditions). I think the unified warning is right (see below), but was a per-arm severity considered for the provably-always-erroring types? Context for why warning looks correct: (1) `apply_filter_to_diagnostics` (`crates/ry-checker/src/diagnostics.rs:397`) overwrites the emitted severity with the registry default, so a per-instance Error here would only be visible to raw API consumers and would reintroduce exactly the divergence this PR removes; (2) the non-literal `character<len=1>` arm has real false-positive potential (R accepts eight string spellings), and the audited posit corpus (`docs/corpus/posit-0.9.0.json`) labels 20 of 22 RY001 findings as false positives — an error-level diagnostic would be unusable in CI. If the always-erroring subsets ever deserve error severity, that would need a separate rule code.
- [nit] `crates/ry-checker/src/tests/type_inference.rs:1955` — the new invariant test covers only the condition family. I audited every `emit()` call site in `ry-checker` at this PR's base: RY001 was the only rule whose emitted severity diverged from `rules::find(code).default_severity`. Since the invariant now holds workspace-wide, a general guard (e.g., asserting emitted severity equals the registry default across the corpus fixtures, or a debug assertion in `Checker::emit`) would prevent the same drift from recurring for other rules.
- [nit] `crates/ry-checker/src/tests/type_inference.rs:1961` — `.find()` only validates the first matching condition-family diagnostic per source. Filtering over all matches and asserting each one would keep the test honest if a source ever produces more than one family code.

Verified beyond the diff: the CLI is genuinely unaffected — the severity filter (`crates/ry-cli/src/check.rs:493`) is applied before the error/warning counts and `exit_code` (`crates/ry-cli/src/check.rs:382`), so `ry check` exit codes and rendered severities were already registry-based and are unchanged; ditto the LSP (`crates/ry-lsp/src/backend.rs:993`), so editor squiggles don't move. The only raw-severity consumers are library API users, the oracle harness (which asserts `Severity::Warning` explicitly at `crates/ry-checker/tests/oracle.rs:340`), and the corpus snapshot — all of which this PR updates; the snapshot diff is exactly the 13 severity lines claimed, and no other `.snap` or test file encodes RY001 severity. The two re-tagged fixtures are the only `must-flag` fixtures whose sole error diagnostic was RY001 (the rest of the `must-flag` set triggers RY040/RY061/RY070/RY080, all still errors), so the oracle update is complete. The `must-warn` tag's requirement that Rscript completes cleanly correctly motivated the `tryCatch`/`stopifnot` error assertions and the removal of the redundant trailing `statement()` call, and the pinned RY002 case (`if (abs(c(1, 2) > 0) > 0)`) matches R semantics (length-2 condition errors) per the ungated first arm in `emit_condition_diagnostics`.

## Overall

Well-scoped fix with the right blast radius: one emit-site change, complete test/snapshot/oracle updates, and new tests that pin both the registry invariant and the RY100/RY002 interaction. The severity choice itself is well-supported by the corpus false-positive data and the filter architecture.

PR #443, opened 2026-09-11
fix(checker): carry compound type guard facts

### LINKED ISSUE #408: Compound || / && guards do not widen default-derived parameter modes (RY061 false positive)
$3

sims1253:
A function such as `f <- function(x = "default") if (is.character(x) || is.null(x)) 1L else x$field` reported RY061 for a caller-supplied list. The character default survived into a branch that rejects character values.

Carry every predicate fact proved on the false path of `||` and the true path of `&&`, including nested negation. Keep those facts local to the branch and preserve the existing scalar/NULL guard rules. The added facts require base predicate resolution and a guard without intervening writes or other calls.

Closes #408.

Validation: workspace tests, Clippy with warnings denied, formatting, and the full R oracle pass. Regression controls cover both branch directions, multiple parameters, rebinding, class predicates, masked functions, and aliases. R accepts the list calls and rejects the atomic control. The fixed 472-package corpus retains all 3,106 baseline diagnostics with no additions or removals; this regression is covered by the focused fixtures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to ry are documented in this file.
 
 ### Fixed
 
+- Widen parameter defaults on the paths proved by compound `&&` and `||`
+  type guards. This avoids false atomic `$` errors for caller-supplied lists
+  while preserving errors on paths the guard does not prove (#408).
+
 - Discard forwarded-default facts after writes before wrapped or nested calls,
   including replacement assignments, loop variables, and literal `assign()`
   targets. Match backticked parameter and argument names consistently
--- a/crates/ry-checker/src/infer/mod.rs
+++ b/crates/ry-checker/src/infer/mod.rs
@@ -712,7 +712,7 @@ impl Checker {
     ) {
         let mut narrowed = HashSet::new();
         let mark = scope.begin_snapshot();
-        if let Some(name) = apply_narrowing_branch(scope, narrowing, NarrowingBranch::Then) {
+        for name in apply_narrowing_branch(scope, narrowing, NarrowingBranch::Then) {
             narrowed.insert(name.to_string());
         }
         for statement in then {
@@ -721,7 +721,7 @@ impl Checker {
         let mut then_delta =
             scope.finish_snapshot(mark, std::mem::take(&mut self.journal_delta_cache));
         let mark = scope.begin_snapshot();
-        if let Some(name) = apply_narrowing_branch(scope, narrowing, NarrowingBranch::Else) {
+        for name in apply_narrowing_branch(scope, narrowing, NarrowingBranch::Else) {
             narrowed.insert(name.to_string());
         }
         if let Some(statements) = else_ {
--- a/crates/ry-checker/src/infer/narrow.rs
+++ b/crates/ry-checker/src/infer/narrow.rs
@@ -12,6 +12,12 @@ use super::*;
 /// from it when that complement is representable.
 #[derive(Debug, Clone)]
 pub(crate) enum Narrowing {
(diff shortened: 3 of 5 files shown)


sims1253 (commented):
This PR teaches the checker to carry conjunction/disjunction type-guard facts through compound conditions: `extract_type_narrowing` now returns a `Narrowing::Compound` that wraps the previous single-fact result plus per-path fact lists (the operands of a true `&&` or a false `||`, with nested `!` flipped through), `apply_narrowing_branch` applies every fact of the taken branch and reports all refined bindings, and a new purity gate restricts compound facts to single-argument, base-resolved predicate calls over identifiers. The facts feed the existing default-parameter degradation machinery, so a parameter like `x = "default"` widens to unknown on paths that prove `!is.character(x)`, fixing the false RY061 from #408; the previous scalar/NULL guard rules are preserved as the compound's `base`.

## Findings

- [minor] `crates/ry-checker/src/infer/narrow.rs:647` — gating compound purity on the strict `resolves_to_base` makes the fix silently unavailable in any file with an attached package: the search-path guard (`resolve.rs:513`) refuses bare-name base resolution whenever `bare_loaded` is non-empty or `search_path_unknown`. I verified empirically that `library(dplyr)` at the top of the file re-introduces the exact false positive this PR fixes (`if (is.character(x) || is.null(x)) 1L else x$field` emits RY061 again), while the single-predicate form `if (is.character(x)) ... else x$field` still degrades in the same file because the builtin path has no provenance check. Consider `resolves_to_base_lenient` here: lexical/parameter shadows, `fn_table`, and imports are still honored by it, and a loaded package redefining `is.character` is about as likely as one redefining `list` (the documented rationale for the lenient variant).

- [minor] `crates/ry-checker/src/infer/narrow.rs:631` — purity is all-or-nothing, so a single non-predicate operand (`is.na(x)`, `length(x) == 1`, a comparison, or a signature-declared predicate like rlang's `is_string`) discards facts proved about *other* variables in the same condition: the false path of `if (is.character(x) || ready)` proves `!is.character(x)` yet keeps no fact, leaving a variant of the #408 pattern unfixed. Sound, but extracting per-operand facts (skipping only the impure leaves) would be a natural follow-up.

- [nit] `crates/ry-checker/src/tests/narrowing.rs:1338` — the false-path controls cover class predicates (`is.data.frame`) and rebinding, but not the `type_is_exactly_tested_family` boundary: `is.numeric(x) || is.logical(x)` with a character default must keep RY061 because the default survives that false path. I verified it currently does; worth pinning as a test since the degradation is what makes the whole fix work.

- [nit] `crates/ry-checker/src/infer/narrow.rs:438` — `apply_narrowing_branch` can return duplicate names when the `base` and a fact refine the same binding; harmless today (`walk_journal_if` folds names into a `HashSet`, the other three callers ignore the result) but slightly surprising for future callers.

Also verified while reviewing: the `&&`/`||`/`!` truth tables are sound, including nested `!(a || b)` and mixed `a || (b && c)` — facts are only taken where R's short-circuit guarantees both operands were evaluated, and `||`-true / `&&`-false paths correctly yield nothing; NA conditions are a non-issue because R errors on an NA condition before either branch runs; facts stay branch-local via the journal snapshots, propagate correctly along diverging-branch continuations (`if (cond) stop() else NULL; x$field` is silent, which is right), and are invalidated by rebinding inside the branch; the fact count is linear in condition leaves, so no state-explosion concern. All 8 compound tests pass on the head commit, and the oracle fixture correctly asserts R rejects the atomic control.

## Overall

A well-scoped, conservative fix with sound fact algebra and good regression coverage; both minors are about where the fix doesn't yet reach rather than where it's wrong.

PR #444, opened 2026-09-11
chore(deps): update MSRV-compatible Rust dependencies

sims1253:
Update serde to 1.0.229, sha2 to 0.11.0, toml to 1.1.6, and flate2 to 1.1.10. This replaces the safe part of #404 while keeping tree-sitter at 0.26.13: tree-sitter 0.27 requires Rust 1.90, above the workspace's Rust 1.88 minimum.

sha2 0.11 also removes the digest output's `LowerHex` implementation. Encode its bytes explicitly so `dump-facts` keeps the same `sha256:` prefix and 64 lowercase hexadecimal digits. Known SHA-256 vectors for empty input and `abc` pin that output.

Validation: `cargo +1.88.0 check --workspace --locked`, workspace tests, Clippy with warnings denied, and formatting pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,7 +128,7 @@ dependencies = [
  "addr2line",
  "cfg-if",
  "libc",
- "miniz_oxide",
+ "miniz_oxide 0.8.9",
  "object",
  "rustc-demangle",
  "windows-link",
@@ -172,11 +172,11 @@ checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
 
 [[package]]
 name = "block-buffer"
-version = "0.10.4"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+checksum = "d2f6c7dbe95a6ed67ad9f18e57daf93a2f034c524b99fd2b76d18fdfeb6660aa"
 dependencies = [
- "generic-array",
+ "hybrid-array",
 ]
 
 [[package]]
@@ -324,10 +324,16 @@ dependencies = [
 ]
 
 [[package]]
-name = "cpufeatures"
-version = "0.2.17"
+name = "const-oid"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ca28b0ae3115b884660db4118d803791fd6756b6e88f39c0f3f7859060d7566"
 dependencies = [
  "libc",
 ]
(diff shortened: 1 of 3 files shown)


sims1253 (commented):
This PR bumps serde to 1.0.229, sha2 to 0.11.0, toml to 1.1.6, and flate2 to 1.1.10 (the MSRV-safe subset of #404, deliberately keeping tree-sitter at 0.26.13 since 0.27 needs Rust 1.90), and adapts to sha2 0.11 removing `LowerHex` on the digest output by encoding the bytes explicitly in a new `format_sha256` helper, pinned by NIST test vectors for empty input and `abc`. I verified the MSRV story against the repo: `rust-version = "1.88"` (Cargo.toml:22) is enforced by a dedicated CI job on the 1.88 toolchain (.github/workflows/ci.yml:20-29), and the RustCrypto 0.11 stack pulled in (digest 0.11.3, hybrid-array, block-buffer 0.12, crypto-common 0.2, const-oid) targets edition 2024 / MSRV 1.85, comfortably under 1.88.

## Findings

- **[minor]** Duplicate-version surface grows in two places: flate2's bump forks `miniz_oxide` (0.8.9 retained for `backtrace`/`object`, 0.9.1 pulled by flate2, Cargo.lock:998), and `serde_derive` 1.0.229 moves to `syn 3.0.2` (Cargo.lock:1543) while the rest of the tree stays on `syn 2.0.118`, so both syn majors now compile. This is unavoidable upstream churn given the PR's goals, but it is a real compile-time/binary-size cost worth accepting consciously — a `cargo tree -d` check on future bumps would keep it visible.

- **[question]** flate2 1.1.10 with default features now compiles *both* `miniz_oxide 0.9.1` and `zlib-rs 0.6.7` (Cargo.lock:502, Cargo.lock:2330) — two DEFLATE implementations in the graph. Upstream docs still describe `rust_backend` (miniz_oxide) as the default and only *plan* to switch to zlib-rs, so the dual pull looks like transition churn rather than something ry chose. If you only want one backend, `flate2 = { version = "1", default-features = false, features = ["rust_backend"] }` (or `zlib-rs` once it is the settled default) would pin that; happy to be told otherwise if this was deliberate.

- **[nit]** The MSRV job runs `cargo check --workspace --all-targets` without `--locked` (.github/workflows/ci.yml:29), so it validates whatever resolution cargo computes rather than the committed lockfile — an inconsistent lock would silently regenerate and pass. Since this PR is exactly about MSRV/lockfile alignment (and the description validated with `--locked` locally), adding `--locked` to that job would lock the guarantee in. Pre-existing issue, not introduced here.

- **[nit]** The changelog has precedent for noting dependency updates (e.g. the thiserror/toml/glob entry around CHANGELOG.md:694) and `[Unreleased]` is currently empty. A one-liner stating that sha2 moved to 0.11 with byte-identical `sha256:` fact output (pinned by tests) would let downstream `dump-facts` consumers confirm stability without diffing the lockfile.

Otherwise this looks solid: sha2 usage is confined to crates/ry-cli/src/facts.rs and both former `{:x}` sites are migrated through `format_sha256` (crates/ry-cli/src/facts.rs:17), the hex table and capacity math are correct, the two pinned vectors are the standard SHA-256 values, only sha2 needed a manifest bump (serde "1", toml "1.1.2", flate2 "1" all already admit the new versions), and `generic-array`/`version_check` drop out of the lock cleanly.

## Overall

A careful, well-scoped MSRV-compatible bump: the one semver-major change requiring code changes is handled correctly and pinned by tests, the lockfile is consistent with the manifests, and CI genuinely exercises the 1.88 MSRV. The only substantive discussion point is whether the duplicate miniz_oxide/syn versions and the dual flate2 backend are an acceptable cost or worth pinning features against.

PR #447, opened 2026-09-11
fix(checker): bind data arguments before masked expressions

### LINKED ISSUE #417: Data-mask follow-ups: TidySelect leading-argument guard, and out-of-order named data misses genuine RY060
$3

sims1253:
Data-mask expressions now use the argument bound to the data formal, even when it appears later in the call. For example, `dplyr::mutate(new = .data$missing, .data = d)` reports RY060 when `d` has a known schema without that column.

The schema and ordinary call paths share the source lookup. Data arguments are checked once in the caller scope. Tidy-select helpers such as `dplyr::vars(1L, .data$column)` keep an unknown mask, and named arguments preserve result schemas and `with()` return types.

Validation: workspace tests, Clippy with warnings denied, formatting, and the full R oracle pass. Tests cover valid and missing columns, absent data bindings, named joins, and quoting helpers. The frozen 472-package comparison is unchanged: 3,106 diagnostics, none added or removed. The comparison caught a valid `vars_rename()` call with a character vector of names; the follow-up keeps its `.data` pronoun opaque and adds an R oracle control.

Closes #417.


sims1253 (commented):
This PR makes the checker resolve a call's data-mask source by formal binding instead of argument position, so a named or trailing data argument (`dplyr::mutate(new = .data$missing, .data = d)`) supplies the mask even when it appears after the masked expressions. `data_mask_source_arg` now falls back to an unmasked first formal when the stub declares no `data_mask_source`, the ordinary call path (`infer/call.rs`) and the schema path (`nse.rs`) share it, joins map `x`/`y` via `match_args_to_params` so named joins get the right schemas, and the `.data` pronoun is only typed as the frame when the mask source is a list/data frame with known columns — keeping quoting helpers such as `vars()` and `vars_rename()`'s character `.vars` opaque (which also removes an RY061 false positive where a leading atomic argument leaked into `.data`). I verified against the vendor stubs that all quoting/masked-first helpers (`vars`, `aes`, `nesting`, `crossing`, `across`, `eval_select`) are correctly rejected by the new guard, that no `schema_effect` signature is rejected by the fallback (so no call loses its schema result), and that R's three-pass matching (including partial names like `.dat = d`) drives the binding.

## Findings

- [question] `crates/ry-checker/src/infer/args.rs:467-484` — the implicit "first formal is the data source" fallback now applies to every resolved signature with any `data_mask`/`tidy_select` eval entry, including ones where the first formal is not the mask data: `recipes::bake(object, new_data, ...)` (true source is `new_data`), `patrick::with_parameters_test_that(desc_stub, code, ...)`, and the `recipes::step_*(recipe, ...)` family. Today this is harmless — those first formals are opaque/character, so the mask degrades to unknown, exactly matching the old arg-0 heuristic — but it is a latent trap: if any of those first-formal types ever gains a column schema, masked arguments would silently check columns against the wrong frame. Since gating the fallback on `schema_effect` would lose stubs like `dplyr::pull`/`tidyr::drop_na` (no `schema_effect`), would declaring `data_mask_source` explicitly on the non-data-first stubs be the safer long-term convention?

- [nit] `crates/ry-checker/src/infer/call.rs:1097-1101` — the data argument is now inferred up front regardless of its source position, so diagnostics from a trailing `data = <expr>` are emitted before diagnostics from earlier arguments (out of span order within one call). This ordering was already true for declared-source formula APIs and is invisible to span-sorted consumers; just noting it now applies to every dplyr/tidyr call site.

- [nit] `crates/ry-checker/src/tests/quoting_data_mask.rs:1473` — coverage is good for binding order, single-check, named joins, and `vars_rename`, but nothing exercises `.env$x` / `{{ x }}` (the `DATA_MASK_ENV_PREFIX` snapshot taken in `dplyr_data_mask_scope`, `nse.rs:141-155`) or `!!`/`:=` injection combined with a named/later data source — reordering the call is precisely when a snapshot-from-base-scope regression would surface.

## Overall

This looks solid: the binding-order fix matches real tidyeval semantics (columns shadow, `.env` and the lexical snapshot bypass the mask), the first-formal guard replaces the old `leading_argument_is_masked` heuristic with a rule I could not break against the stub corpus, and the frozen 472-package comparison plus the new R oracle controls give strong regression confidence. Only the latent fallback-heuristic question above is worth a decision; everything else is polish.

Issue #380, opened 2026-09-08
formals(f) <- alist(...) closures analyzed as static source (distr6, 17 FPs); suppress empty-name RY010 from recovered parse trees


sims1253:
## Summary

Two smaller capability/hygiene gaps from the corpus RY010 triage, filed together because both concern dynamically-constructed code ry analyzes as static source:

1. **`formals(f) <- alist(...)` builds functions whose formals/bodies ry never sees** — distr6 constructs S4 coercion methods this way (17 RY010 instances): the generated closures reference names provided only through the `alist`/`body<-` machinery, and the *construction site* itself is clean R.
2. **RY010 with an *empty* variable name fires on unparseable files** — 5 corpus instances where a file also emits RY000 (recovered parse tree) and the unbound-variable analysis then reports `variable `` is not bound`. Diagnostics derived from a recovered tree under a known-parse-failure should be suppressed (the parse failure already covers the file).

## Evidence

ry 0.9.0, 472-package corpus:

- distr6 (and R6-heavy packages) define method templates like `formals(fun) <- alist(x =, ...)` then `body(fun) <- substitute(...)`; calls through the resulting closures are flagged on names that exist only in the construction context. 17 instances, all FP.
- Files with RY000 "unparseable region (recovered tree may be unreliable)" additionally emit RY010 with empty backticks — visible in the corpus JSON as `variable `` is not bound in this scope`.

## Suggested direction

- For (1): treat closures whose `formals<-`/`body<-` are assigned dynamically as opaque (silence RY010/RY080 on their internals rather than analyzing the placeholder), and/or record the `alist` names at the construction site as bindings in the constructed function's scope.
- For (2): when a file's parse produced a recovered tree (RY000 present), suppress downstream semantic diagnostics for that file — the parse diagnostic is the actionable signal.

## Context

Found in the ry 0.9.0 corpus audit round 2, RY010 cluster triage (artifacts `audit-20260908-ry090-r2/audit-results/ry010/`, mechanisms M15 and the empty-name flag).

sims1253:
**0.9.1 release screening: Optional / later.**

Split empty-name recovery hygiene (optional) from dynamic formals/body analysis (later). Prefer suppressing invalid recovered regions over all semantic checks in a file.

**Scheduling:** outside the 0.9.1 milestone. It can be admitted before scope freeze once the bounded change has a reproducer, adjacent valid controls, and time for review.

[...]

PR #466, opened 2026-09-15
chore(msrv): raise the Rust minimum to 1.90 and update tree-sitter to 0.27

sims1253:
## Summary

- Raise the workspace MSRV from 1.88 to 1.90: tree-sitter 0.27.0 and tree-sitter-language 0.1.8 require rustc 1.90 (per the MSRV failure on #465), and this unblocks the parser update that has slipped for three cycles (#390, #404, #465).
- Re-pin the CI MSRV job to the dtolnay/rust-toolchain 1.90.0 branch tip (49fda20), which hardcodes toolchain 1.90.0, and update the building-from-source docs in docs/usage.md.
- Cherry-pick dependabot's tree-sitter 0.26.13 to 0.27.0 cargo-dependencies bump from #465 (commit keeps dependabot authorship). Verified locally: cargo check --workspace clean.

Closes #465 (superseded: same dependency update lands together with the MSRV floor it requires).

## Summary by CodeRabbit

* **Documentation**
  * Updated build instructions to require Rust 1.90 or newer.

* **Chores**
  * Updated the minimum supported Rust version to 1.90.
  * Upgraded the tree-sitter dependency to version 0.27.
  * Adjusted workspace configuration and CI verification accordingly.

PR #467, opened 2026-09-15
fix(checker): suppress semantic diagnostics for files parsed from recovered trees

sims1253:
## Approach

When a file's parse produced a recovered tree, the pass-3 walk still ran
every semantic rule over parser-repaired structure. The 0.9.0 corpus audit
found 5 files where RY010 fired with an empty variable name (`variable `` is
not bound in this scope`) on top of a parse error: statements, calls, and
identifiers in a recovered tree can be artifacts that exist nowhere in the
source, so those findings are noise on top of the actionable signal, which
is the parse failure itself.

The suppression hooks into `Checker::emit_diagnostics`
(crates/ry-checker/src/lib.rs), the single seam every diagnostic path
funnels through (`check`, `check_with_scope`, and the multi-file
`Project` per-file emitters, covering CLI check and the LSP alike):

- RY000s are emitted first, exactly as before, and still fail the run.
- The semantic walk still runs, so scope records, reference facts (which
  already carried a `parse_error` blocker), and assignment-type capture
  keep serving editor features on files that are merely mid-edit.
- Afterwards, if the file has any parse error, everything the walk
  emitted is dropped. Clean files are untouched.

## Test evidence

- Unit tests in crates/ry-checker/src/tests/diagnostics.rs for both
  directions: a broken file reports only RY000s; a clean file keeps its
  RY010.
- New corpus fixture testdata/err_recovered_tree_suppresses_semantics.R
  with an exact `# expect: RY000` multiset (the source used to emit
  RY010 + RY000).
- New oracle fixture testdata/oracle/recovered_tree_semantics_suppressed.R
  (must-flag: R rejects the incomplete header; ry reports RY000 as the
  only finding).
- Snapshot deltas triaged: the corpus snapshot's err_parse_error.R entry
  loses its RY010 `f` finding (3 RY000s remain), and the editor
  playground's broken.R loses RY010@21:11 (`daily_shots`, an artifact
  of the recovered tree) -- README table and prose refreshed. The glue
  and purrr vendor snapshots are unchanged (no vendored file has RY000).
- CHANGELOG and the RY000 row in docs/rules.md updated.

Closes part 2 of #380 (recovered-tree suppression); the
`formals(f) <- alist(...)` opacity half (the distr6 17-FP batch) stays
open for a follow-up. Refs #380

## Summary by CodeRabbit

- **Bug Fixes**
  - Files with syntax errors now report only parse-error diagnostics (RY000); unreliable semantic warnings are suppressed.
  - Clean files continue to report semantic diagnostics, such as undefined-variable warnings.

- **Documentation**
  - Updated rule guidance and editor examples to reflect diagnostic behavior for files with syntax errors.

- **Tests**
  - Added coverage confirming semantic diagnostics are suppressed for recovered parse trees.


sims1253:
Post-hoc review (merged as 5031e861 before the review bots and any human review finished; see adjudication below). I verified the merged commit directly in a detached worktree.

## Verdict

**Approve with follow-ups.** The implementation is correct, matches the part-2 scope of #380, passes every gate I could run, and the "single seam" claim holds under inspection. No P0/P1. The follow-ups are granularity and harness questions, not defects against the stated intent.

## Findings

### P2 — File-level suppression drops genuine semantic errors in clean regions (intended, but a real trade-off to own)

`crates/ry-checker/src/lib.rs:1308-1310`. Suppression keys on "any parse error in the file" and truncates everything the walk emitted. Verified empirically: a file consisting of `print(undefined_thing)` + two clean statements + a broken `f <- function( {` tail emits `[RY000, RY000, RY000]` on the merged commit, but `[RY000 x3, RY010 x2]` on the parent (839ed72) — the genuine `undefined_thing` RY010 is sacrificed along with the parser-repair artifact. A 99%-valid file with one broken region now reports nothing semantic, including for statements entirely before the break.

Defenses that make this acceptable: (1) #380's suggested direction explicitly asks for per-file suppression ("suppress downstream semantic diagnostics for that file"); (2) R itself refuses the whole file — `parse()` on such a file fails at the first break, so no semantic claim about the file is actionable until it parses (confirmed with Rscript); (3) the codebase already treats such files as untrusted elsewhere: reference facts carry a whole-scope `parse_error` blocker (`reference_facts.rs:143-155`), RY097 suppresses all diagnostics for probably-not-R files, and `dump-facts` refuses parse-error files outright (`ry-cli/src/facts.rs:319`); (4) the walk still runs, so scope/reference/assignment captures keep serving hover, completion, hints, and rename mid-edit.

Follow-up worth a deliberate decision: region-aware suppression (e.g. drop only diagnostics at or after the first ERROR/MISSING span, or spanning into it) would preserve the `undefined_thing` case. It is harder than it sounds — tree-sitter repair restructures everything after the break (the `x <- function(` case invents a call over the rest of the file), so "first error onward" is the plausible refinement, and it should be validated against the 5 corpus instances before adopting. Note the truncation also drops error-severity semantic diagnostics (e.g. RY060), not just warnings; `docs/rules.md`'s "reports only its RY000s" is the accurate phrasing.

### P2 — Suppression correctness now depends on grammar fidelity

`crates/ry-core/src/parser.rs:1018-1056` (`collect_parse_errors`: tree-sitter `ERROR` nodes and `MISSING` tokens, nothing else). If the tree-sitter grammar ever flags valid R as an error, this change silently silences all diagnostics for that valid file, where before it merely added an RY000. This is guarded today (parser corpus asserts `parse_errors.is_empty()` on valid sources, real-world and vendor snapshot tests all pass), but the blast radius of a future grammar regression is strictly larger now. Worth remembering when bumping tree-sitter grammars (#466 bumped to 0.27 immediately before this).

### P3 — Oracle harness cannot pin "RY000 and nothing else"

`crates/ry-checker/tests/oracle.rs:347` — `must-flag` only asserts at least one Error diagnostic, so the new fixture (`oracle/recovered_tree_semantics_suppressed.R`) proves the R-side premise (R errors) and that ry flags, but the "semantic rule must stay silent" half is pinned only by the corpus fixture and unit tests, which never consult R. Consider a `must-flag-only RY000` tag (or extending `must-warn` semantics) if this suppression class grows. Also noting the new fixture correctly omits `# oracle-claim:` — RY000's claim is already registered by `syntax_error_claim.R`.

### P3 — Mid-edit LSP UX confirmation

While typing, an editor user now sees only RY000 squiggles until the file parses again; semantic diagnostics vanish the moment syntax breaks and return when it repairs. This is what #380 asked for and is consistent with the capture-preserving design, but it is a louder behavior change for editor users than for CLI users — flagging for conscious sign-off rather than as a defect.

### Verified non-findings (skeptical checks that came back clean)

- **Single seam claim holds.** `check` and `check_with_scope` both funnel into `emit_diagnostics` (`lib.rs:1039, 1050`); Project pass 3 builds per-file emitters that call it (`project.rs:786`); the LSP `publish_diagnostics` routes through `Project::check_with_workspace` (`ry-lsp/src/backend.rs:940-960`). Pass 1 is silent by design and pass 2 cannot emit (`discarding` guard, `resolve.rs:567`); `emit` is the only `diagnostics.push` site. Nothing can precede the RY000s in the vector, so `semantic_start` cleanly splits parse from semantic.
- **No emit path escapes the truncation.** RY097 (CLI, `check.rs:455`) fires only on files excluded from checking entirely; total parse failures are skipped with a stderr note (`check.rs:417-428`); `dump-facts` refuses parse-error files (`facts.rs:319`). Exit-code semantics unchanged (RY000 is error-severity; `check.rs:391` already failed on parse errors).
- **Tests pin both directions.** `recovered_tree_suppresses_semantic_diagnostic` (all-RY000) and `recovered_tree_suppression_leaves_clean_files_alone` (RY010 kept). Corpus fixture uses `# expect: RY000`, which is an exact multiset in `corpus.rs`.
- **Snapshot delta is exactly the intended removal.** `err_parse_error.R` loses only the RY010 `f` entry (span 154-155, line 3); the three RY000s are byte-identical; the only other snapshot change is the new fixture's entry. The editor playground README delta (RY010@21:11 `daily_shots` removed, warnings 1 -> 0) matches.
- **Conventions.** Conventional commit subject; no emojis; comment style matches CONTRIBUTING ("explain R semantics and design constraints"); CHANGELOG `### Fixed` under `[Unreleased]` in the fix commit is established practice (#454 bb00f99, #455 e735d1e) — not a deviation; `docs/rules.md` update appropriate.

## Bot-comment adjudication

| Bot | What it posted | Ruling |
| :--- | :--- | :--- |
| pullfrog[bot] (17:48:42Z) | Unchecked progress checklist only; its `create_pull_request_review` step never ran (no review exists in the reviews API) | Nothing to adjudicate — bot did not complete before the merge |
| coderabbitai[bot] (17:48:55Z, updated 17:50:48Z) | "Review failed — The pull request is closed"; walkthrough summary only, zero findings | Nothing actionable — the run aborted because the PR was merged underneath it; this failure is a direct consequence of the premature merge, not a comment on the code |
| Socket / Macroscope | No comments or reviews on this PR (checked issue comments, review comments, reviews, and the timeline) | N/A — nothing present |

Net: zero substantive bot findings exist on this PR. This comment is the first actual review the change received, which is the process problem — merging ahead of both configured bots forfeited the review safety net entirely.

## Local verification

Merged commit 5031e861 in a detached throwaway worktree; direct rustup toolchain binaries.

- `cargo test --workspace` — PASS (exit 0)
- `cargo test -p ry-checker --test oracle` — 16 passed, 0 failed (1 ignored)
- `cargo test -p ry-checker --test oracle -- --include-ignored` — 17 passed, 0 failed (full R-executed matrix, includes the new fixture)
- `cargo clippy --workspace --all-targets -- -D warnings` — PASS
- `cargo fmt --all -- --check` — PASS on the committed tree
- Oracle premise: `Rscript --vanilla` on `recovered_tree_semantics_suppressed.R` exits 1 with "unexpected end of input"; `parse()` on the corpus fixture fails whole-file at 8:0 — `must-flag` premise holds, and the whole-file failure is the strongest argument for file-level granularity
- Scratch test (throwaway, deleted; run on both parent 839ed72 and merge 5031e861): mixed file `print(undefined_thing)` + clean defs + broken `function(` tail — parent `[RY000 x3, RY010 x2]`, merged `[RY000 x3]`; clean control keeps RY010 on both. Suppression confirmed and trade-off characterized exactly.

PR #473, opened 2026-09-15
ecosystem: reconcile posit ledger with recovered-tree suppression

sims1253:
## What disappeared, and why it is intended

PR #467 (issue #380, part 2) suppresses semantic diagnostics on files whose parse produced a recovered tree, keeping the RY000 syntax error as the actionable signal. The posit drift gate has failed on every PR since, because seven reviewed RY010 findings recorded in `docs/corpus/posit-0.9.0.json` no longer appear in the hermetic reports:

- lintr `tests/testthat/dummy_packages/RConfigInvalid/lintr_test_config.R:3:1`
- testthat `tests/testthat/test-parallel/syntax-error/tests/testthat/test-error-1.R` at 5:1, 5:5, 5:13, 5:15, 5:22, and 6:1

This PR reconciles the ledger with that delta. The gate is not weakened: `reconciliation: hermetic` stays, all identities remain pinned, and any future missing or unowned identity still fails.

## Per-finding verification

Both corpus files were fetched at their pinned commits (lintr `990e578`, testthat `9b6f12b`) and fed to release binaries built from either side of the #467 merge:

- **pre-#467** (`839ed72`): lintr's file emits RY000 3:1 + RY010 3:1 (`variable \`\` is not bound`); testthat's file emits all six RY010s (`but`, `this`, `a`, `syntax`, `error` at 5:1/5:5/5:13/5:15/5:22 and the empty name at 6:1) + RY000 6:1 — exactly the seven ledger identities.
- **post-#467** (`5031e86`, current main): both files emit **only** their RY000; every RY010 is gone.

So each disappearance traces directly to the recovered-tree suppression, not to an unrelated regression, and the RY000 identities remain in the reports (the failing job did not flag them missing).

## Ledger and report updates

- `docs/corpus/posit-0.9.0.json`: the seven findings removed (398 -> 391); `classification.false_positive` 361 -> 354; `audit_group_counts` test-fixture 73 -> 71, type-narrowing 57 -> 52; package `diagnostics` lintr 9 -> 8, testthat 8 -> 2; `source_sha256` recomputed over the pruned `posit.*.root.txt` set; `ry_commit` moved to the #467 merge; a note records the delta.
- `docs/corpus/posit-messages-0.9.json`: the same seven identities pruned (396 -> 389).
[...]

Issue #371, opened 2026-09-08
New rule: length()/sum() applied directly to a comparison — length(x == y) parenthesization (2 shipped TPs: shiny, packrat)


sims1253:
## Summary

`length(x == y)` — aggregation applied to a *comparison* instead of comparing the aggregation — is the `any(x) == 0` parenthesization family applied to `length()`. Two shipped true positives found in the corpus: one in shiny (shipped for years), one in packrat.

## Evidence

Both from the ry 0.9.0 corpus run and verified as real (but note: ry found these *correctly* via RY033 — this issue is about the rule family for future precision work, and these serve as proven regression fixtures):

1. **shiny `R/bookbook-state.R:138`** — `if (length(inputVals != 0))` where line 118 has the correct `length(inputVals) != 0` form nearby. `length(<logical>)` is always ≥ 1 when any elements exist, so the condition is effectively always-TRUE; benign-in-effect typo, 3 corpus copies of the file.
2. **packrat `inst/vendor/renv.R:7108`** — `length(package == 1)` is always 1; intended `length(package) == 1`.

The general shape: `length(x == y)`, `length(x > n)`, `sum(x == y)`-style aggregations whose direct argument is a comparison — mechanically detected, high precision, and the sibling of #356 (`any(x) == 0`) and the closed #322 grep-length family.

## Suggested direction

Warning rule: `length()` (or `sum()`/`mean()` over logicals) applied *directly* to a comparison expression — almost always a misplaced parenthesis (`length(x) == y` intended). The two shipped sites above are ready-made true-positive fixtures; guard tests should keep `length(x == y)` legitimate uses silent where a length is genuinely intended (rare; usually in comments/docs).

## Context

Found in the ry 0.9.0 corpus audit round 2 (artifacts `audit-20260908-ry090-r2/audit-results/operators/`, true-positive section).

sims1253:
**0.9.1 release screening: Later; rewrite scope.**

Exclude legitimate sum/mean-of-comparison idioms; investigate a narrow length-parenthesis heuristic only.

**Evidence and scope correction:**

#371 is unsafe as written: `sum(x == y)` is ordinary counting and `mean(x == y)` is an ordinary proportion. #356 also needs valid boolean-to-number comparison controls. Keep these as heuristic design proposals, not quick high-precision wins.

[...]

PR #476, opened 2026-09-15
feat(checker): flag scalar comparisons on any()/all() results

### LINKED ISSUE #356: New rule: comparisons on any()/all() results (any(x) == 0 parenthesization)
## Summary

`any()`/`all()` return a length-1 logical; comparing that scalar with `== 0`, `> 1`, or similar is almost always a parenthesization mistake (`any(x == 0)` intended). A shipped defect in glue demonstrates the pattern; ry 0.9.0 has no rule for it.

## Evidence

glue @ da9c73f, `R/utils.R:32` in `recycle_columns()`:

```r
if (any(lengths) == 0) { ... }     # intended: any(lengths == 0)
```

Runtime-verified: `recycle_columns(list(a = character(0), b = c("x", "y", "z")))` errors "Variables must be length 1 or 3" instead of returning `character()`. The same commit (099b283, "0 length inputs produce 0 length outputs") added the correct `any(lengths == 0)` form twice in `R/glue.R:139,191`, making the intent unambiguous. Currently masked in glue's only call path by a pre-filter, so latent rather than user-visible today.

## Suggested direction

Warning rule: a comparison operator whose operand is a direct `any(...)`/`all(...)` call — the result is a scalar logical, so numeric comparison (`== 0`, `> 0`, `!= 1`, ...) is either always-FALSE/always-TRUE or a misplaced parenthesis. Keep it to direct call positions (no dataflow) for precision. An RY093/RY100-style parenthesization heuristic fits the existing rule architecture.

## Context

Found in the ry 0.9.0 tidyverse corpus audit (artifacts `audit-20260908-ry090/audit-results/glue/`). The prior Posit corpus study's A1 (`"name" <- value` in `list()`) is a sibling "mechanical mistake" rule family.

sims1253:
Closes #356.

## Approach

RY107 `any-all-scalar-comparison` (warning): a comparison operator with a direct base-resolving `any(...)`/`all(...)` call (exactly one positional argument) on one side and a numeric literal on the other, **when the comparison does not preserve the call's scalar logical value**. `any()`/`all()` return a length-1 logical that numeric comparison coerces to 0/1, so the outcome is exactly one of three families, decided by evaluating the operator at FALSE=0 and TRUE=1:

- **negating** (`== 0`, `!= 1`, `<= 0`, `< 1`): computes `!any(x)` — reported as `TRUE exactly when the call is FALSE`.
- **constant** (`> 1`, `>= 2`, `< 0`, `== 2`, `>= 0`, ...): dead guard — reported as `always FALSE`/`always TRUE`.
- **preserving** (`== 1`, `!= 0`, `> 0`, `>= 1`): computes exactly what the bare call computes — **silent**.

The preserving carve-out is the crux. The original `constant-condition` sketch was rejected (see the former recall.rs module doc) because "is always FALSE" was wrong (`FALSE == 0` is TRUE) and because the shape seemed indistinguishable from diffobj's legitimate `!all(diff(x)) == 1L`, pinned must-stay-silent in `testdata/ry095_ry096_real_shapes.R`. Classifying by **outcome** rather than shape separates them exactly: diffobj's comparison preserves the value, glue's negates it. NA input propagates to NA in every family, changing neither classification. The message suggests the element-level rewrite with mirrored operands (`0 == any(x)` suggests `any(x == 0)`).

## Neighboring-rule positioning map

| shape | rule |
|---|---|
| `length(x == y)`, `nchar(x == y)` — comparison INSIDE a counting call | RY093 |
| `abs(x > y)` — comparison inside a numeric math call | RY100 |
| `length(any(1L)) > 0` — `length()` of a scalar reduction vs 0 | RY105 |
| `any(x == 0)` — comparison inside any()/all() | nothing (correct spelling) |
| **`any(x) == 0` / `all(x) > 1` — comparison OUTSIDE with any()/all() as direct operand** | **RY107 (new)** |

No span is double-reported: RY093/RY100 are emitted at the enclosing call before its arguments are inferred, so RY107 (running at the comparison's BinOp site) scans for a same-span RY093/RY100 and yields — the same suppression pattern `emit_condition_diagnostics` already uses for RY100 vs RY001/RY003. RY105 requires a `length()` call as an operand, a different shape from the bare any/all call, so the two cannot meet on one comparison. Shadowed callees (`any <- function(...)`), foreign-qualified ones (`otherpkg::any`), multi-argument calls (`na.rm`), and non-literal operands stay silent; `base::any` keeps the premise.

## R-verified evidence (Rscript, R 4.6.1)

- `length(any(c(1,2)))` is 1, `typeof` logical; `any(c(NA,FALSE))` is NA and `NA == 0` is NA.
- `FALSE == 0` is TRUE / `TRUE == 0` is FALSE (negating); `TRUE > 1` and `FALSE > 1` both FALSE (constant); `TRUE == 1L` TRUE / `FALSE == 1L` FALSE (preserving, the diffobj family).
- glue divergence: with `lengths <- c(0L, 3L)`, `any(lengths) == 0` is FALSE while `any(lengths == 0)` is TRUE — the written guard skips the zero-length branch (the issue's `recycle_columns` error).

## Fixtures and tests

- `testdata/err_any_all_scalar_comparison.R` — the glue repro plus the adjacent quiet idioms (element-level spelling, diffobj shape, `== 1`, `sum(x > 0)`, `na.rm` form); `# expect: RY107`.
- `testdata/oracle/any_all_scalar_comparison_claim.R` — `must-warn RY107` + `oracle-claim`, R-side stopifnot assertions of the premise.
- `testdata/err_recall_rules_repro.R` A7 un-marked as "deliberately missed"; `tests/recall_rules.rs` pins fire/silent directions, mirrored suggestion, shadowing, and the neighbor-yield cases (8 new tests); unit tests in `src/tests/diagnostics.rs` (6); probe in `tests/probes.rs`; verdict + R7 case in `tests/rule_evidence.rs` (consistent); `docs/rules.md` and `docs/corpus/rule-evidence-0.9.md` rows.

## Ecosystem reconciliation (second commit)

Hermetic reruns of both manifests: RY107 adds exactly one identity everywhere — glue `R/utils.R:32:7`, the audited defect itself — recorded as true_positive in both ledgers (`upstream-glue` / `upstream-package`); summaries and the posit message ledger updated; tidyverse `source_sha256` recomputed over the fully regenerated set. The posit fast tier also surfaced **pre-existing drift from #467** (merged after the last posit reconciliation): seven stale false-positive RY010 identities in syntax-error fixtures (lintr, testthat) whose semantic diagnostics the recovered-tree suppression now withholds by design; their RY000s remain. Removed with explanation in the ledger note. `clean_checkout.rs` was fixed to relativize identities against the package root like `run.sh`'s writer (`R/utils.R` instead of a bare file name) — the empty baseline had masked that the two formats disagree.

## Gates

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (60 suites), `cargo test -p ry-checker --test oracle` (plus `--include-ignored`), both `ecosystem/run.sh --check` gates (tidyverse + posit fast tier), and the reconciliation/drift/manifest-isolation/label-falsification integration tests all pass locally. Vendor snapshot: one new hit, the glue true positive, triaged in the test's comment block.

## Summary by CodeRabbit

* **New Features**
  * Added warning RY107 for suspicious comparisons between `any()`/`all()` results and numeric literals.
  * Flags comparisons that negate the logical result or always produce a constant outcome.
  * Preserves intentional value-based comparisons such as `== 1`, `> 0`, and `!= 0`.

* **Documentation**
  * Added RY107 to the rule reference, changelog, diagnostic summaries, and ecosystem reports.

* **Tests**
  * Added coverage for flagged patterns, valid comparisons, corrected element-wise forms, and interactions with existing rules.


sims1253:
## Verdict: approve-with-followups

RY107 is semantically correct (verified against R 4.6.1 across the full operator/literal matrix), precise on real idioms (zero false positives in 12 corpus-plausible shapes), correctly positioned against RY093/RY100/RY105, and the ledger work is exactly the claimed one-identity delta in each ledger. One P3 convention gap (CHANGELOG) plus P3 observations. The tidyverse "regeneration" this PR was asked to be adjudicated on is **not in this PR at all** — evidence below.

## Findings

**P3 — CHANGELOG `[Unreleased]` has no RY107 entry** (`CHANGELOG.md:5-13`). A new user-facing warning rule belongs under `### Added` per the 0.10.0 precedent. The rest of the conventions check passes: `docs/rules.md` row present, rule-evidence row and completeness counts updated (35 rows / 34 claim fixtures, enforced by `rule_evidence.rs`), `ry explain-rule RY107` resolves, oracle claim fixture present (`testdata/oracle/any_all_scalar_comparison_claim.R`, `must-warn RY107` + `oracle-claim`), conventional-commit subjects (`ecosystem:` type has repo precedent in #473), no emojis in the diff, the merge commit is a clean reconciliation (delta vs `4982750` is exactly two corpus docs).

**P3 — negative-literal constant cases stay silent** (`crates/ry-checker/src/infer/recall.rs:87-93`). `numeric_literal` matches only `Expr::Integer`/`Expr::Double`; `-1` parses as unary minus over `1`, so `any(x) > -1` (always TRUE), `any(x) < -1` (always FALSE), `== -1`, `!= -1` do not fire (CLI-verified). No negating form is affected (negation requires a literal in [0,1)), so this is a constant-family recall gap only. The helper is shared with RY105, which has the same limit — consistent family behavior, low real-world frequency. Follow-up material, not a blocker.

**P3 (note, deliberate) — logical-literal spellings out of scope.** `any(x) == FALSE` computes the same negation as `== 0` but stays silent; pinned in `ry107_stays_silent_against_non_literal_operands` and documented in the PR. Acceptable; revisit only if corpus evidence surfaces.

**P3 (pre-existing, not this PR — for the record).** `docs/corpus/tidyverse-0.7.1.json` contains two exact-duplicate dbplyr RY091 entries (`tests/testthat/test-backend-.R:110:36` and `tests/testthat/test-translate-sql-string.R:13:36`, each appearing twice). Present on main since before this branch forked; `check-ledger.py` totals still agree because the duplicates are counted on both sides. Also `docs/corpus/README.md`'s digest-recipe paragraph still says the six non-block packages' reports "contain 46 unowned findings in total"; the actual count is 44 since #460's typeshed 0.5.1 refresh emptied scales' report. Both fit a small corpus-housekeeping follow-up.

## Classification re-probe matrix (R 4.6.1 via `Rscript --vanilla`, plus built CLI at 19165ab)

Premise confirmed: `any()`/`all()` return a length-1 logical; comparisons coerce FALSE/TRUE to 0/1; `any(c(NA,FALSE))` is NA and every comparison propagates NA (`NA == 0`, `NA > 1`, `NA <= 1` are all NA), so a comparison's family is fully determined by its outcomes at 0 and 1, and NA changes no classification.

| op | lit | at FALSE / at TRUE | family | CLI behavior |
|---|---|---|---|---|
| == | 0 | T / F | negating | fires, suggests `any(x == 0)` |
| == | 1 (also `1L`, `1e0`) | F / T | preserving | silent |
| == | 2, 0.5 | F / F | constant FALSE | fires |
| != | 0 | F / T | preserving | silent |
| != | 1 | T / F | negating | fires |
| != | 2, 0.5 | T / T | constant TRUE | fires |
| < | 1, 0.5 | T / F | negating | fires |
| < | 0, -0.5 / 2 | F/F, T/T | constant | fires |
| <= | 0, 0.5 | T / F | negating | fires |
| <= | 1, 2 | T / T | constant TRUE | fires ("always TRUE") |
| > | 0, 0.5 | F / T | preserving | silent |
| > | 1, 2, 1.5, `1e0` | F / F | constant FALSE | fires |
| >= | 1, 0.5 | F / T | preserving | silent |
| >= | 0 | T / T | constant TRUE | fires |
| >, >=, ==, != | -1 | constant | constant | silent (P3 above) |
| `0 == any(x)` | 0 | T / F | negating | fires, mirrored suggestion `any(x == 0)` |

Every fired/silent decision matches the R-derived family; no misclassification found (no preserving form fires; no negating or constant form is silent, apart from the negative-literal P3). Two review-prompt expectations corrected against real R: `<= 1` is **not** value-preserving (`FALSE <= 1` and `TRUE <= 1` are both TRUE — constant TRUE) and the rule correctly fires on it; `!= 2` is constant TRUE and fires. Runtime divergence re-verified: with `lens <- c(0L, 3L)`, `any(lens) == 0` is FALSE while `any(lens == 0)` is TRUE.

## Precision and positioning (built CLI at 19165ab)

- 12 corpus-plausible idioms, zero RY107: `any(grepl("a", x)) == FALSE`, `isTRUE(any(x))`, `any(x) == n` (variable), `any(x) == any(x)`, `sum(x > 0)`, `sum(x == 0) > 0` (#371's family), `any(x == 0)`, `all(x != 1)`, `any(x, na.rm = TRUE) == 0`, `any(x) == FALSE`, `!all(diff(x)) == 1L` (diffobj), `any(x) == TRUE`.
- Qualification/shadowing: `base::any(v) == 0` fires; `otherpkg::any(v) == 0` silent; locally shadowed `any` silent; `%in%` and `== NA` silent (RY034 owns NA); fires in `while()` conditions too.
- Same-span dedup: `length(any(x) == 0)` reports RY093 only; `abs(any(x) == 0)` RY100 only; `nchar(any(x) == 0)` RY093 only; `length(any(1L)) > 0` RY105 only; a combined `length(...)+abs(...)` line reports RY093+RY100 with no RY107 and neither neighbor swallowed. Mechanism verified in source: RY093/RY100 emit at the inner comparison's span from the enclosing-call check (`crates/ry-checker/src/infer/call.rs:856,986`) before the comparison's BinOp arm, where RY107 scans for the same-span neighbor (`recall.rs:400-404`).
- Flagship: the built CLI on the vendored glue checkout yields exactly one diagnostic across all 12 files — `R/utils.R:32:7` RY107 suggesting `any(lengths == 0)` — and `R/glue.R:139,191` carry the intended spelling.

## Regeneration adjudication

**Ruling: nothing to remediate in this PR — the tidyverse catch-up regeneration is not here.** Git archaeology:

- This branch forked from main at `5031e86` (#467). At that commit and at the PR base `07e67b1` (#473), `tidyverse-0.7.1.json` already contained 74 findings (10 TP / 41 FP / 23 unowned).
- The 103-to-74 drop (29 unowned ggplot2 RY010 identities) landed on main in `1784245` ("ecosystem: reconcile reports and ledgers with the 0.5.1 vendor"), part of merged release PR #460, before this branch existed. The 0.10.0 CHANGELOG entry states it outright: "Resolving the scales inventory clears the ggplot2 RY010 false-positive batch from both committed ecosystem ledgers."
- This PR's net ledger delta vs main is exactly one identity per ledger: glue `R/utils.R:32:7` RY107 true_positive (tidyverse 74-to-75, TP 10-to-11; posit 391-to-392, TP 37-to-38), plus the README summary rows catching up to the #460-era ledger values (the rows had been stale on main: 103 / +52 vs actual 74 / +23).
- All 29 dropped identities spot-checked at the pin: I ran the built branch binary (identical to main for these shapes — no RY107 fires on ggplot2; the clean_checkout change is test-only) on the cached ggplot2 checkout at the exact manifest pin `6870419aa6e1...`: zero of the 29 locations fire. The dropped lines reference scales-package names (`rescale`, `squish`, `censor`, `rescale_max`, `fullseq`, `ContinuousRange`) that the r-typeshed 0.5.1 scales inventory now resolves — consistent with #460's attribution.

On the fold-in question as posed: folding a full ledger regeneration into a feature PR would have deserved pushback (#473 precedent stands), but this PR contains no such regeneration — only the single-identity delta its own reconciliation produced, and a two-line README row alignment that the repo convention already owes any ledger-touching PR. Appropriate as merged; no split-out needed. If the review record believed a tidyverse catch-up was folded in here, correct the record, not the PR.

## clean_checkout.rs scope ruling

**Keep — appropriate fold-in, and a forced one.** The old test flattened diagnostic paths to `file_name()` while `run.sh`'s `write_report` relativizes against the package dir (verified: `ecosystem/run.sh` strips `package_dir/`, producing `R/utils.R:32:7`), so the two formats could never agree on a non-empty report. The committed `glue.txt` has been empty, making the comparison vacuously green; RY107 makes it non-empty for the first time, so without this fix the baseline test fails on this very PR. A standalone fix on main would be untestable there (empty vs empty passes either way, demonstrable only via the falsification test). The fix mirrors run.sh's relativization with a `file_name()` fallback — separately motivated but causally coupled; carrying it here was the right call.

## Bot adjudication

- **pullfrog initial review** (posit `source_sha256` stale after the PR edited posit root reports; README rows lagging both ledgers): valid, verified fixed in `19165ab`. Recomputed `cat ecosystem/reports/posit.*.root.txt | sha256sum` = `a88205e71ba3...` matching the committed ledger value; README rows now match both ledgers (38/354/0 at 392; 11/41/0 plus 23 unowned at 75).
- **pullfrog incremental review** ("no new issues; no code came via the merge; notes not duplicated; digests verified"): verified accurate. `git diff 4982750 19165ab` touches exactly `docs/corpus/README.md` and `docs/corpus/posit-0.9.0.json`; the #473 seven-removal note is preserved once with a distinct RY107 note appended; `ry_commit` records the full 40-char `9c1e3c17e4fb...` in both ledgers.
- **coderabbit**: rate-limited at review time; rechecked after local verification — still no findings posted. Nothing to adjudicate.

## Local verification (detached worktree at 19165ab)

- `cargo test --workspace`: all suites pass, 0 failures.
- `cargo test -p ry-checker --test oracle -- --include-ignored`: 17 passed, 0 failed (includes the new claim fixture's R-side `stopifnot` assertions).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `python3 ecosystem/check-ledger.py docs/corpus/posit-0.9.0.json docs/corpus/tidyverse-0.7.1.json`: OK (392 = 38 + 354; 75 = 11 + 41 + 23; package sums agree).
- Digests: posit recipe `a88205e7...` matches; tidyverse recipe (concatenated non-posit `*.root.txt`) `3e3699c3...` matches.

Suggested non-blocking follow-ups: add the CHANGELOG `[Unreleased]` entry; a small corpus-housekeeping pass for the duplicate dbplyr RY091 entries and the README "46 unowned" sentence; optionally fold unary minus into `numeric_literal` for the negative constant family (RY105 would benefit too).

PR #481, opened 2026-09-15
feat(checker): flag self-referential formal defaults (RY109)

### LINKED ISSUE #364: Diagnose self-referential formal defaults (copy = copy, j = j) — shipped dtplyr bug ry missed
## Summary

A formal whose default expression references itself (`copy = copy`, `j = j`) is a guaranteed runtime error when the default is triggered ("promise already under evaluation: recursive default argument reference"), but ry emits nothing even for trivial cases. This was a real shipped bug in dtplyr.

## Evidence

demonstrated on dtplyr's parent commit bffe46e (the audited HEAD, dbe32a6, is literally titled "Fix recursive helper defaults" and fixes exactly this with regression tests):

```r
# R/step-join.R:162 (pre-fix)
dtplyr_auto_copy <- function(x, y, copy = copy, ...) { ... }

# R/tidyeval-across.R:6,20 (pre-fix)
quo_expand_across <- function(quo, j = j) { ... }
```

Runtime (R 4.6.1): calling without the argument errors `promise already under evaluation: recursive default argument reference`. ry 0.9.0 emits nothing on the pre-fix source (verified during the audit on that blob).

The direct form is not always the bug — a default like `x = x` where the intent is "fall back to the *caller's* variable of the same name" is a real (if dubious) R idiom that works when the caller supplies a same-named variable... actually no: within the function, `x = x` in formals always refers to the promise itself and always recurses when triggered. The only legal variant is referencing a *different* formal.

## Suggested direction

High-precision lint (error-level, when the default is a bare symbol equal to the formal's own name): `function(x, y = y)` and `function(copy = copy)`. Slightly broader second tier: defaults referencing the same formal through one level of expression (`n = n + 1`) — same recursion, still mechanical to detect. R's parser hands us the formal list; no inference needed. Guard tests: dtplyr's fixed call sites and `stats::qqnorm(formula)`-style legitimate cross-formal references must stay silent (only *self*-references flag).

## Context

Found in the ry 0.9.0 tidyverse corpus audit (artifacts `audit-20260908-ry090/audit-results/dtplyr/`; the pre-fix repro and ry silence were verified against the parent commit).

sims1253:
## What

Closes #364. A formal whose default expression references the formal itself (`copy = copy`, `j = j`, `n = n + 1`, `caller_env = caller_env()`) can only resolve to its own promise. New rule RY109 (`self-referential-default`, warning) flags them; RY098 keeps its stricter proven-forcing half of the diagnosis unchanged, and RY109 covers the rest.

## R-verified semantics (Rscript --vanilla, R 4.6.1)

- Triggering the default errors: `promise already under evaluation: recursive default argument reference or earlier problems?` — for the bare form, the compound form (`n = n + 1`), and the callee form (`x = x()` / `tz = tz(x)`, where the formal shadows a real same-named function, so the intent "call the enclosing function" cannot work).
- **Supplying the argument is fine** (`f(copy = 1L)` never touches the default), and a promise that is never forced or is defused before forcing also runs (`function(x = x) 1L`, `enquo(x)` capture, `x <- 1L` replacement). Verified each shape.

## Decision: warn unconditionally on the default itself, at warning severity

The rule warns whenever the default self-references in an executed position and the body does not provably force the promise (RY098 owns the provably-forced cases; the union is unconditional). Rationale:

- A self-referential default can never *evaluate* successfully — whether this body happens to force it only decides when the bug bites. dtplyr shipped exactly the unproven-forcing shape: conditional or lazily-forwarded forcing was invisible to RY098's proof, which is why ry 0.9.0 was silent on the pre-fix blob.
- The 0.9.1 screening note (#364 comment: "avoid unconditional new errors ... `function(x = x) 1L` succeeds") is honored on both counts: the severity is warning, not error, and RY098's forcing analysis is untouched.
- The defusing idiom is respected, not just tolerated: a body that hands the promise to a reviewed capture helper stays quiet (see precision below). What remains possible is a defused-by-an-unprovable-helper default warning — hence warning severity.

This decision is pinned in tests: `err_recursive_default_lazy_forward.R`, `err_recursive_default_defused_forward.R`, `err_recursive_default_unforced_branch.R`, and `never_used <- function(copy = copy) 1L` in `err_self_referential_default_unproven.R`.

## Precision boundary (deliberate idioms stay quiet)

- Defaults referencing a *different* formal (`x = y, y = 1L`) are legal R and never flag (fixture `ok_non_self_referential_defaults.R`, also pinning dead-branch defaults `if (FALSE) x else 3L`).
- Quoted defaults that capture the formal (`x = quote(x)`, `substitute(x)`, `expression(x)`, `alist(x)`) stay quiet — the predicate is RY098's capture-aware evaluation-order walk (`first_executed_identifier`).
- A body that defuses the promise with a reviewed capture helper — `rlang::enquo`/`enexpr` (`captures_promise`), `substitute`/`quote`/`expression` (`quoted_expression`), or tidy injection `{{ x }}` — uses the shape deliberately and stays quiet. A bare defuser name *defined by the project itself* is not credited without provenance (a local `quote <- function(x) x` is not a defuser). Fixtures: `ok_recursive_default_captured.R`, `ok_recursive_default_qualified_defuser.R`, `err_recursive_default_shadowed_defuser.R` for the negative side.
- Mutual recursion between defaults (`x = y, y = x`) also errors in R when every cycle member is missing (verified), but detecting it needs a formal-reference cycle analysis; left open, noted in `ok_non_self_referential_defaults.R`.

## Evidence

- dtplyr parent commit bffe46e: ry flags exactly `R/step-join.R:162` (`copy = copy`) and `R/tidyeval-across.R:6,20` (`j = j`); the fix commit dbe32a6 (PR tidyverse/dtplyr#501, "Fix recursive helper defaults", `copy = FALSE` / `j = TRUE`) is RY109-clean.
- Oracle: `oracle/self_referential_default_claim.R` (`must-warn RY109` + `oracle-claim: RY109`) demonstrates in R that the dtplyr-shaped conditional-force helper errors when the argument is missing and returns normally when supplied.
- Vendored purrr: `as_progress(..., caller_env = caller_env())` — runtime-verified latent true positive (formal shadows rlang's `caller_env()`; forcing errors; all internal callers currently supply the argument). Triaged in `vendor_snapshot.rs`.
- Corpus fixtures: the dtplyr repro plus lazy/defused/never-used/short-circuit/replaced/missing() shapes now expect RY109; RY098's forced shapes (`err_recursive_parameter_default.R`, force-contract fixtures) are unchanged.

## Corpus deltas (rebased on #472, both manifests regenerated and strict-gated)

- tidyverse: +10 findings, all true positive (dtplyr x3, purrr x1, ggplot2 x2, lubridate x2, dplyr x1, dbplyr x1) — 77 -> 87 diagnostics, 13/41 -> 23/41 TP/FP (+23 unowned unchanged).
- posit: +25 findings — 23 latent true positives (gt, shiny, sparklyr, pkgdown closure-capture misconceptions; httr/purrr/pkgdown callee-shadowing; ellmer/infer/parsnip/shiny eager forwarding; the shared tidyverse set) and 2 false positives on rlang's own defusing tests (`test-nse-defuse.R:329,334`, where bare `enexpr`/`enquo` are defined by rlang itself and cannot be credited without provenance — warning severity keeps the cost low). 396 -> 421 diagnostics, 42/354 -> 65/356.
- Suppressed as deliberate idioms (no finding): corrr's `x = x`/`y = y` enquo + `{{ }}` capture, dbplyr's `sql_runif(n = n())` whose body verifies the captured default via `enquo`.

## Gates

`cargo fmt --all -- --check`; `cargo clippy --workspace --all-targets -- -D warnings`; `cargo test --workspace` (61 suites); `cargo test -p ry-checker --test oracle -- --include-ignored`; `ecosystem/run.sh --check` for both manifests; `check-ledger.py` on both ledgers. Ledger/report edits are isolated in the second commit for an easy union with main.

## Summary by CodeRabbit

* **New Features**
  * Added the RY109 warning for self-referential function defaults that can fail when evaluated.
  * Distinguishes safely captured or defused defaults from defaults that may trigger promise-evaluation errors.
  * Expanded RY098 messaging for cases where a self-referential default is provably forced.

* **Documentation**
  * Added rule-reference documentation, changelog entries, and ecosystem reports covering RY109 findings.

* **Tests**
  * Added broad coverage for recursive defaults, conditional execution, lazy forwarding, capture helpers, and shadowing scenarios.


sims1253:
## Independent review: RY109 self-referential formal defaults

### Verdict

**Approve with comments.** No blockers. Every load-bearing claim reproduced under independent verification: the R semantics, the corpus identities at their ledger pins, the ledger arithmetic and digests, and RY098's non-regression. The findings below are three documentation-accuracy errors in the ledger notes/docs, two suppression-side false-negative divergences between R and the `body_defuses_formal` model that should at least be pinned as known gaps, and one missing doc row. Nothing invalidates a `true_positive` label.

### Findings

**P2 — defuse credited on one branch while another branch forces the formal (false negative).**
`f <- function(x = x, flag) if (flag) rlang::enquo(x) else x` — R 4.6.1 errors on `f(flag = FALSE)` ("promise already under evaluation"; verified), while `f(flag = TRUE)` runs defused. ry is silent on this file (reproduced with the release CLI). `body_defuses_formal` (`crates/ry-checker/src/infer/quoting.rs:813-863`) credits any reviewed capture call or `{{ x }}` anywhere in the frame, regardless of sibling bare uses of the formal. Suggested refinement: refuse the credit when the formal also appears in a non-capture position in the same frame (that would fix this shape; the walk already sees both). At minimum, pin the current behavior in a fixture comment the way the over-warning direction is pinned.

**P2 — defuse-then-force is a false negative and is undocumented.**
`f <- function(x = x) { q <- rlang::enquo(x); rlang::eval_tidy(q) }` errors in R (verified with real rlang installed; base twin `q <- substitute(x); eval(q)` likewise errors). ry is silent (reproduced with the CLI). The PR text acknowledges only the over-warning direction of the defusing credit ("a defused-by-an-unprovable-helper default warning"); the under-warning direction — capture credited, capture later evaluated — is real, is not acknowledged anywhere, and is not fixable by the sibling-read refinement alone (no bare read of `x` remains). Document as a known gap of the suppression, ideally with a `known-gap`-style comment or fixture.

**P2 — the posit ledger note overclaims what was runtime-verified.**
`docs/corpus/posit-0.9.0.json` note says the 23 true positives are "each runtime-verified: ... errors with R's 'promise already under evaluation' whenever the argument is missing." That is false for at least three identities: dplyr `R/distinct.R:81` and dbplyr `R/verb-distinct.R:61` (`caller_env = caller_env(2)` — the formal is never consumed by either body, so the call **runs** when it is missing; verified) and ggplot2 `R/stat-bindot.R:133` (`method = method` — `densitybin(1:3)` runs; the formal is unused in the body). For these, the verified fact is "the default can never evaluate", which is exactly how the tidyverse note phrases it ("a default that can only resolve to its own promise"). The TP labels remain defensible for all three — the defaults are dead on arrival and would error the moment anyone consumes them — but the posit note's blanket sentence should be corrected, and the RY109 message text ("errors ... when the argument is missing", `quoting.rs:169-175`) shares the same overstatement for never-forced formals; "cannot evaluate; errors if the promise is ever forced" would be accurate in both places.

**P3 — the tidyverse ledger note contradicts the reports and the posit note on rlang.**
`docs/corpus/tidyverse-0.7.1.json` note: "Deliberate defusing idioms stay silent: ... rlang's test-nse-defuse.R fixtures ...". They do not stay silent — `ecosystem/reports/rlang.root.txt`, which is part of this manifest's digest, gains exactly `tests/testthat/test-nse-defuse.R:329:22` and `334:22` in this PR, and the posit ledger labels those same two `false_positive`. Two smaller inaccuracies in the same note: corrr is a posit-corpus package, not in this manifest; and "`error_call = error_call()`" attributed to the dplyr/dbplyr pairs is pkgdown's shape (`R/topics.R:271`) — dplyr's `error_call = caller_env()` (`distinct.R:82`) is a sibling-formal reference that correctly stays quiet and is not flagged.

**P3 — `docs/corpus/rule-evidence-0.9.md` is missing the RY109 row.**
RY106 and RY107 have rows; `rule_evidence.rs` was updated (R7_NA_RULES, VERDICTS) but the markdown table drifted because the tests only enforce the in-code tables. Add the row (23 TP / 2 FP posit, 10 TP / 0 FP tidyverse, claim fixture `self_referential_default_claim.R`, verdict keep).

**P3 — bare-defuser crediting depends on import resolvability; worth documenting.**
Bare `enquo(x)` is credited when rlang is resolvable (`library(rlang)` or package `NAMESPACE` `import(rlang)` — corrr verified) and flagged in a standalone file with no import (reproduced). Reasonable behavior, but it is precisely why rlang's own tests warn, and one sentence in the RY109 docs entry would save users the confusion.

### Corpus triage audit (R 4.6.1; sources fetched at the ledger pins; minimal triggering calls)

| Category | Identity | R verdict when default triggered | Ruling |
| :-- | :-- | :-- | :-- |
| shipped bug | dtplyr `R/step-join.R:162` `copy = copy` (bffe46e) | errors, recursive-default | TP; ry flags exactly this pre-fix; fix commit dbe32a6 is RY109-clean (both verified by local CLI runs) |
| shipped bug | dtplyr `R/tidyeval-across.R:6` `j = j` | errors (across_setup forces `j`) | TP |
| shipped bug | dtplyr `R/tidyeval-across.R:20` `j = j` | errors | TP |
| callee-shadowing | purrr `R/progress-bars.R:56` `caller_env = caller_env()` | errors when the stop_input_type branch is reached | latent TP; all internal callers (map.R:201, map2.R:72, pmap.R:134) supply it — verified at pin |
| callee-shadowing | httr `R/config.r:189` `config = config()` | errors immediately (stopifnot forces) | latent TP |
| callee-shadowing | pkgdown `R/topics.R:271` `error_call = error_call()` | errors | latent TP |
| callee-shadowing | dplyr `R/distinct.R:81`, dbplyr `R/verb-distinct.R:61` `caller_env = caller_env(2)` | **runs** — formal never consumed | TP defensible (default can never evaluate); note wording overclaims (see P2 finding) |
| closure-capture | pkgdown `R/tweak-navbar.R:32` `pkg = pkg` (nested `get_hrefs`) | errors | latent TP |
| closure-capture | gt `R/cols_align.R:197-198` `data_tbl`/`col_classes` | errors | latent TP |
| closure-capture | gt `R/utils_render_grid.R:421` `data = data` | errors | latent TP |
| closure-capture | sparklyr `R/connection_test.R:2-3` `master`/`config` | errors (list() forces) | latent TP |
| eager forwarding | ellmer `R/provider-claude-files.R:149` `type = type` | errors (curl::form_file forces) | latent TP |
| eager forwarding | infer `R/calculate.R:741` `call = call` | errors (error path) | latent TP |
| eager forwarding | parsnip `R/aaa_models.R:483` `call = call` | happy path runs; error path errors | latent TP, correctly so (error-path default) |
| eager forwarding | shiny `R/bootstrap-layout.R:609` `width`/`height` | errors | latent TP (fillRow/fillCol always supply) |
| eager forwarding | ggplot2 `R/stat-quantilemethods.R:81` `method.args = method.args` | errors (inject `!!!method.args`) | latent TP |
| eager forwarding | ggplot2 `R/stat-bindot.R:133` `method = method` | **runs** — formal never consumed | TP defensible; note wording overclaims (P2 finding) |
| eager/callee | lubridate `R/parse.r:975` `tz = tz(x)` | errors on the character branch | latent TP |
| eager/callee | lubridate `R/guess.r:265` `.select_formats` | errors (body calls the shadowed function) | latent TP |
| false positive | rlang `tests/testthat/test-nse-defuse.R:329` | runs: `fn()` returns `quote(x)` (real rlang) | correctly labeled FP |
| false positive | rlang `tests/testthat/test-nse-defuse.R:334` | runs: returns two identical quos (real rlang) | correctly labeled FP |

**Ruling on the two acknowledged FPs.** Reproduced and confirmed as false positives. The implementer's rationale is sound: crediting a project-defined bare defuser without provenance would also credit a test-local `quote <- function(x) x`, which forces and errors (verified in R), so blanket crediting is unsafe. But rlang is the canonical defusing package and warnings on its own defusing tests are a bad look forever. Ruling: acceptable documented FP at warning severity for this PR; file a follow-up refinement — credit an in-project defuser name when the project is provably the definer (package identity via DESCRIPTION/NAMESPACE export) or when the in-project definition is provably an alias/wrapper of a reviewed qualified defuser. Both would clear these two FPs without opening the shadowing hole.

**Claimed suppressions verified.** corrr (aa0488f): `R/retract.R` `{{ val }}`/`{{ x }}` injections and `R/reshape.R` `stretch_unique` bare `enquo(x)` credited via `NAMESPACE` `import(rlang)` — no RY109 on the full package (local CLI run). dbplyr `sql_runif` `n = n()` (`R/translate-sql-scalar.R:139`, body `quo_get_expr(enquo(n))`): runs in R with `n` missing (verified with a substitute twin) and stays quiet in a full `R/` run that reproduces the exact single dbplyr ledger identity.

### RY098 non-regression check

Clean. The refactor at `quoting.rs:153-176` evaluates the same conjunction as before — `first_executed_identifier` and `guaranteed_force_before_replacement` are both pure, only their evaluation order moved — so RY109 fires in exactly the `RY098-did-not` residue. Confirmed empirically: `git diff origin/main` over `ecosystem/reports/` is purely additive (74 insertions, 0 deletions, 46 files) with **zero** RY098 lines touched in the reports or in the fixture snapshots; the converted fixtures' headers explicitly pin "RY098 must stay silent".

### Bot adjudication

CodeRabbit: rate-limited, produced no findings — nothing to adjudicate. Pullfrog: posted an unchecked progress checklist only, no findings. Rechecked once after local verification; still nothing.

### Local verification

- Detached worktree at b80fe9d. `cargo test --workspace` pass; `cargo test -p ry-checker --test oracle -- --include-ignored` 17 passed (R-executed, ~35 s); `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean.
- `check-ledger.py`: OK for both ledgers. Independently recomputed from the JSONs: tidyverse 87 = 23 TP / 41 FP / 23 unowned (10 RY109, all TP; audit groups sum 87); posit 421 = 65 TP / 356 FP (25 RY109: 23 TP + 2 FP; every per-package `diagnostics` matches its findings count; groups sum 421). Both `source_sha256` digests reproduce byte-exact from the 32 non-posit and 62 posit root reports in filename order. SUMMARY RY109 rows sum correctly (tidyverse 10 owned / 12 root; posit 23 owned / 25 root).
- CLI hand-runs at pins: dtplyr bffe46e flags exactly the three ledger identities; dbe32a6 clean; dbplyr f478e20 exactly `R/verb-distinct.R:61`; corrr aa0488f silent.
- Conventions: conventional commits (`feat(checker): ...`, `docs(corpus): ...`), no emojis in the diff, oracle fixture carries `# oracle: must-warn RY109` + `# oracle-claim: RY109` and is R-executed by the gate, `docs/rules.md` row present, CHANGELOG entry present, RY098 summary-row update is an accurate clarification of pre-existing behavior, probe positive/negative wired. 14 `ok_` fixtures were converted to `err_`/replaced (5 detected as renames, 9 delete+new), each pinning the unconditional-warn decision with RY098-silence pinned in the header.
- Union expectations for the merge coordinator: on top of this PR, #478 adds exactly one tidyverse TP (hms `R/hms.R:307` RY108, verified `true_positive` in its diff) — expect tidyverse 88 = 24 / 41 / 0 (+23 unowned) and posit unchanged at 421; #480 touches no corpus or ledger files. Note all three PRs append at the same insertion points (RULES, PROBES, VERDICTS, rules.md, CHANGELOG, SUMMARY tables, corpus snapshot), so the unions will conflict textually while remaining semantically additive.

sims1253:
Review follow-ups pushed in c0031d1 + 58a414d. Thank you for the audit -- every item folded in, nothing re-triaged.

**P2 wording overclaim (fixed in message + ledger).** RY109 now reads "has a self-referential default that can never evaluate; forcing the promise errors ('promise already under evaluation')" (quoting.rs, with a comment naming the three never-consumed identities). The posit note replaces the blanket "errors whenever the argument is missing" with the split you suggested: most identities verified to error on a triggering call; dplyr distinct.R:81, dbplyr verb-distinct.R:61, and ggplot2 stat-bindot.R:133 verified as dead-on-arrival defaults (the call runs when missing). Goldens regenerated: corpus and purrr vendor snapshots, posit message ledger. No identity moved; 87/421 hold and both digests are byte-identical.

**P2 defuse-then-force (pinned).** `oracle: known-gap` fixture `self_referential_default_defuse_then_force_gap.R` (enquo then eval_tidy, R errors uncaught, ry silent), plus a Known misses block in body_defuses_formal's doc comment, and the PR description's defusing paragraph now names both directions.

**P2 branch divergence (pinned, not implemented -- evidence below).** `oracle: known-gap` fixture `self_referential_default_branch_divergence_gap.R` (also asserts flag=TRUE returns a quosure before the flag=FALSE error). I tried the suggested sibling-non-capture-read refinement and it does not fall out cleanly, for two reasons grounded in the verified corrr suppressions: (1) after `x <- enquo(x)`, later bare reads see the replacement quosure, not the promise -- corrr stretch_unique does exactly `y <- enquo(y); ... as_label(y)`, so a sibling-read refusal re-flags the deliberate idiom; (2) corrr retract's `stats::xtabs(val ~ x + y, data)` references are bare only without formula/data-mask context. A sound version needs stop-at-reassignment flow analysis plus quoting-context tracking; that rationale is recorded in the module comment so a future refinement has the requirements written down.

**P3s (all done).** rule-evidence-0.9.md gains the RY109 row (33/2, claim fixture, keep, known-gap pins noted). The tidyverse note now attributes `error_call = error_call()` to pkgdown (posit), calls dplyr's `error_call = caller_env()` the legal sibling it is, moves corrr to a posit-corpus mention, and states rlang's fixtures warn (2 posit FPs; the ungated rlang root report in this manifest gains the same two identities). docs/rules.md gains the bare-defuser crediting sentence: a bare helper is credited only when it resolves via an attached package or NAMESPACE import, which is also why rlang's own tests warn.

Gates re-run on the result: fmt, clippy -D warnings, cargo test --workspace (61), oracle --include-ignored (17 passed; both gaps registered, not stale), run.sh --check both manifests, check-ledger both. Also took the follow-up you suggested for the rlang FPs as a candidate refinement (credit an in-project defuser when the project provably defines it via its own NAMESPACE export, or when the definition is provably an alias of a reviewed qualified defuser) -- not attempted here to keep this PR's surface confined to review fixes.

Issue #486, opened 2026-09-16
lsp: R source changes never refresh the disk index; closing a saved file restores the last indexed snapshot


sims1253:
## Problem

The language server treats changes to unopened `.R` files as non-events, and closing an edited-and-saved document falls back to the **stale indexed snapshot** instead of current disk content. Subsequent checks analyze old source and produce stale cross-file diagnostics until some unrelated event (config/DESCRIPTION/NAMESPACE/native/serialized-data change) triggers a reindex.

## Evidence (at d45e3e9)

- `refresh_watchers` registers globs only for `ry.toml`, `DESCRIPTION`, `NAMESPACE`, `src/*.{c,cc,cpp,cxx}`, and `*.{rda,RData,rdata,json}` — no R source pattern (`crates/ry-lsp/src/backend.rs:543-549`).
- `did_change_watched_files` early-returns for anything else, even if a client forwards `.R` events: `if !resolution_changed { return; }` (`crates/ry-lsp/src/backend/handlers.rs:387-389`).
- `did_close` removes the doc, versions, parse, tree, and project-cache entries (`handlers.rs:400-443`) but never re-reads the file; `state.disk_files` is only ever written by `spawn_background_index` (`backend.rs:1168`).
- `publish_diagnostics` merges `disk_files` entries for non-open paths (`backend.rs:872-884`); `disk_files` holds parsed `SourceFile` snapshots with no mtime/freshness check. There is no `did_save` handler, and the publish path performs zero disk I/O by design (`tests/protocol_contract.rs` asserts this).

So after "edit → save → close" (with another doc open), the closed file is re-analyzed from its last-background-index snapshot — typically the initialize-time content.

Related but distinct: #81 (close/reopen convergence, fixed) and #206 (custom config paths unwatched, fixed). The state-machine harness works around this gap by faking `ry.toml` events for `.R` create/delete/rename (`tests/session_state_machine.rs:628-666`), and its model never saves buffers to disk, so save-then-close is untested.

## Expected

Every persisted source transition (unopened-file edit/create/delete, save-then-close) converges with a fresh server over the same final filesystem; current open text always overrides disk.

## Fix direction

Register the supported R source extensions as watched globs and process their events through an index update/invalidation path; on `did_close`, refresh the disk entry for that path (reusing the bounded decoder and generation checks — never replacing a live open buffer with a disk read).

## Tests

[...]

PR #519, opened 2026-09-17
Clear diagnostics for every document that leaves the analysis set

### LINKED ISSUE #489: lsp: disabling or excluding multiple documents leaves stale diagnostics on all but one URI
## Problem

When documents become ineligible (folder disabled, files excluded), they are removed from the next check but do not all receive empty diagnostic publications. Old findings stay visible in the editor. Closed disk-indexed files that drop out of the publication set after a rescan have the same gap.

## Evidence (at d45e3e9)

- `publish_diagnostics` sends an empty publication only for the **requested** URI when it is ineligible (`crates/ry-lsp/src/backend.rs:857-861`); other newly-ineligible documents are simply filtered out of `doc_versions` and `disk_entries` and never published again. There is no reconciliation against the set of previously published URIs (no such tracking exists in `ry-lsp`).
- Debouncing makes it worse: `schedule_diagnostics` uses one workspace-wide generation (`backend.rs:1228-1247`), so scheduling N URIs — which is exactly what `republish_all_open_documents` does (`backend.rs:1211-1219`) — leaves only the last task alive; the other N−1 abort as stale.
- Concrete path: disable a folder (or exclude two files) with both docs open → `did_change_configuration` → ... → `republish_all_open_documents` (`handlers.rs:249`). Two schedules bump the generation twice; only the second task runs, clears its one URI; the other keeps its old squiggles indefinitely. Which URI survives depends on `docs` HashMap iteration order. The mixed case (one file excluded, one still eligible) is worse: the excluded one may never be rescheduled at all.
- Closed disk files: after a rescan drops a file from `disk_files` (deletion, newly-excluded), nothing publishes an empty set for its URI either.
- Existing mitigations cover only adjacent cases: `did_close` clears its own URI (`handlers.rs:433-436`) and folder removal clears URIs under removed roots (`handlers.rs:343-357`). The disable/exclude path has no equivalent.
- Tests only cover the single-document case, where the surviving task's requested URI *is* the affected file (`tests/code_action_eligibility.rs:64-101`, `tests/configuration_refresh.rs:203-244`). The state-machine test also filters empty publications out of comparisons and has no disable/exclude operation.

## Expected

Every URI that previously received diagnostics and is removed from the check by disablement, exclusion, folder removal, or deletion gets an empty publication — for all of them, not just the last-scheduled one.

## Fix direction

Track the set of URIs with published diagnostics per partition and reconcile it against each successful current-generation result, sending empty publications for the set difference; keep it generation-safe. At minimum, `publish_diagnostics` could clear every open document failing `eligibility_for_path`, not just the requested URI.

## Tests

At least two simultaneously affected documents (so the single-requested-URI behavior can't pass): open and closed indexed files, folder disable, multi-file exclusion, folder removal, deletion followed by rescan, and re-enablement. Assert the client's final diagnostic store, not the latest publication for one document.

sims1253:
## Summary

Two gaps combined into #489, both verified on main with the new tests failing:

- **Debounce starvation**: `schedule_diagnostics` used one workspace-wide generation, so scheduling N URIs — exactly what `republish_all_open_documents` does after a settings/config change — left only the last task alive; the other N−1 aborted as stale, and which URI survived followed `docs` HashMap iteration order.
- **No reconciliation**: `publish_diagnostics` sent an empty publication only for its *requested* URI when that URI was ineligible. Every other document that left the eligible set (folder disabled, file excluded) was filtered out of the check and never published again, and closed disk files a rescan dropped from `disk_files` (deleted, newly excluded) had the same gap — nothing tracked which URIs had received diagnostics before, so their old squiggles lingered indefinitely. The existing mitigations covered only adjacent cases: `did_close` (own URI) and folder removal (open docs under removed roots).

The fix keeps the debounce's burst-collapsing purpose and adds a tracked-publication set:

- Every scheduled URI joins `State::pending_diag_paths`; the one surviving task drains the whole set atomically with its staleness check and publishes it, so a batch of scheduled URIs all gets its pass (`publish_diagnostics` now takes the drained set; a requested-but-ineligible URI still gets its immediate empty acknowledgment, which `didOpen` into a disabled folder relies on).
- `State::published_paths` records URIs whose last publication carried diagnostics. Each pass updates it from what it actually sent, then `clear_dropped_diagnostics` publishes empty sets for tracked URIs that left the analysis set: no longer eligible (folder disabled, excluded), or neither open nor indexed (closed file deleted or dropped by a rescan). It decides against *current* state rather than one pass's snapshot, so a slower older pass that interleaves after a newer one cannot leave a cleared URI dirty — its own trailing reconciliation re-clears.
- The reconciliation also runs where no open document can drive a debounced pass: `republish_all_open_documents` with zero open docs, and `did_change_workspace_folders` when every open document lived under removed roots. `did_close` and folder removal untrack the URIs they explicitly clear.

## Test plan

New `crates/ry-lsp/tests/diagnostic_reconciliation.rs` session tests (all three verified to fail on main across 5 runs; each keeps ≥2 documents in play so the surviving task's own URI cannot pass, and asserts the client's whole final diagnostic store via `quiesce_diagnostics`, where the last publication per URI wins):

- `disabling_the_folder_clears_every_open_document` — disable with both docs open clears **both** URIs, and re-enabling restores both (proving a publication, not permanent silencing).
- `excluding_one_of_two_open_documents_clears_only_the_excluded_one` — the mixed case: the excluded URI is cleared while the still-eligible one republishes its finding.
- `closed_disk_files_dropped_by_a_rescan_get_an_empty_publication` — a never-opened indexed file cleared after deletion (rescan via the watched-`ry.toml` channel) and after newly excluding it, while the open driver document keeps its diagnostics.

The state-machine convergence property needed no filter change: its empty-publication filter models "a URI published with `[]` is equivalent to an absent URI", and the fix only adds empty publications for URIs that carry no diagnostics on a fresh server either; `pr_session_seeds` passes unchanged.

Local gates: `cargo test -p ry-lsp` green (153 passed), `cargo fmt --all -- --check` clean, `cargo clippy -p ry-lsp --all-targets -- -D warnings` clean.

Closes #489

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale diagnostics remaining visible when workspace folders are disabled or files are excluded.
  - Ensured all affected open documents receive cleared diagnostics, rather than only the last processed document.
  - Cleared diagnostics for closed files removed during rescans, including files deleted from disk or newly excluded.
  - Cleared diagnostics for closed files under workspace folders that are removed.
  - Improved diagnostic restoration when a workspace folder or previously excluded file becomes eligible again.


sims1253:
On pullfrog's informational finding about closed-file deletions only reconciling when a watched event drives a rescan: the missing `.R` watcher registration is a pre-existing detection gap, deliberately left out of scope for this PR. It is tracked by issue #486, which is being implemented separately.

PR #520, opened 2026-09-17
fix(lsp): apply size and depth eligibility gates to open documents

### LINKED ISSUE #488: lsp: open documents bypass the documented max-file-bytes and max-depth eligibility gates
## Problem

A file omitted by bounded background discovery (over `index.max-file-bytes`, or below a pruned `index.max-depth`) enters project-wide analysis as soon as it is opened. The open-document eligibility gate enforces excludes and folder enablement only — not the size/depth limits — contradicting the documented editor contract, and the admitted buffer's definitions leak into every other open file's resolution.

## Evidence (at d45e3e9)

- `State::eligibility_for_path` checks folder enablement and `ry_workspace::is_file_eligible` (`crates/ry-lsp/src/backend.rs:453-470`).
- `is_file_eligible` is excludes-only; its signature cannot even express a depth (`crates/ry-workspace/src/discovery.rs:10-30`). Size/depth caps live only in the directory walker: depth pruning at `discovery.rs:355-359`, size check at `discovery.rs:392-399` — applied only to files admitted through `discover_r_files`.
- Open documents are admitted through that same excludes-only gate (`backend.rs:840-852`) and parsed with no size guard into `project_files` → the shared `Project` (`backend.rs:863-884`). Code actions and inlay hints gate identically.
- `docs/configuration.md:110-113`: "In the editor, open documents that are ineligible for analysis (excluded by `exclude` patterns, over `max-file-bytes`, or below a pruned `max-depth`) ... do not enter project-wide binding or diagnostic state." Only the exclude third is implemented.

## Repro

Set a small `max-file-bytes` (e.g. 32) with an ordinary source file above it defining a unique binding. Index with the file closed, then open the identical file unchanged: its binding now resolves for other files. Repeat with a file below a directory level pruned by a small `max-depth`.

## Fix direction

Share the eligibility policy between bounded discovery and open-buffer admission, with two nuances: use the **buffer text length**, not `fs::metadata` (unsaved pasted content can also cross the boundary), and compute depth as path depth relative to the folder root, since walk depth is not a path property. Keep lightweight editor features (highlighting) working for ineligible open documents; keep them out of project-wide state.

## Tests

Closed-to-open parity for both limits, edits crossing the size boundary, configuration changes restoring eligibility, and cleanup of previously published diagnostics. Don't treat membership in an old disk index as a substitute for evaluating the current buffer.

sims1253:
## Summary

A file omitted by bounded background discovery (over `index.max-file-bytes`, or below a pruned `index.max-depth`) entered project-wide analysis as soon as it was opened: `State::eligibility_for_path` enforced `exclude` patterns and folder enablement only, so the admitted buffer's definitions leaked into every other open file's resolution, contradicting the documented editor contract.

Eligibility now runs through one shared policy — `ry_workspace::is_file_eligible_with_limits` — that the directory walker and open-buffer admission both call. Two nuances, as the issue prescribes:

- the size gate measures the **open buffer's text length**, not `fs::metadata`: unsaved pasted content can cross the boundary without touching disk. Closed paths pass `None` and skip the size gate (the walker already measured them on admission);
- the depth gate counts the **containing directory's components relative to the folder root**, since walk depth is not a path property.

Ineligible open documents stay out of binding and diagnostic state while keeping single-document editor features: inlay hints (computed from the document's own parse, no project state) keep serving them; `enable: false` still returns null there. Code actions stay gated on eligibility. Two subtler points: every open path — not just the eligible ones — shadows its disk twin, so a buffer that grows past `max-file-bytes` cannot keep contributing its path through the stale small on-disk twin the index still holds; and edits or configuration changes crossing a boundary eject or re-admit the buffer, with stale diagnostics clearing through the shared dropped-URI reconciliation.

`docs/configuration.md` already promises this behavior, so no doc change was needed.

## Evidence

- `crates/ry-workspace/src/discovery.rs`: `is_file_eligible` is excludes-only; the size cap (`metadata.len() > max_file_bytes`) and depth pruning (`depth >= max_depth`) live only in `discover_recursive`. The new `is_file_eligible_with_limits` combines all three, with the walker kept as the reference implementation.
- `crates/ry-lsp/src/backend.rs`: `eligibility_for_path` now calls the shared policy with the buffer length and the folder root; `publish_diagnostics` snapshots only eligible open docs into the project and every open path shadows its disk twin.
- `crates/ry-lsp/src/backend/handlers.rs`: the inlay-hint gate is narrowed to the `enable` switch (document-local analysis needs no project state).
- Repro from the issue, now pinned as tests: with `max-file-bytes = 32`, indexing a 41-byte `def.R` (defining the character `f`) omits it while closed, and opening the identical buffer leaves `use.R`'s `f() + 1L` call clean (no RY040). Same for `max-depth = 1` with `a/b/def.R`.

## Test plan

New protocol suite `crates/ry-lsp/tests/open_document_eligibility.rs` (7 tests, cross-file shadowing observable `f() + 1L` → RY040 iff the character `f` participates):

- closed-to-open parity for the size gate (identical over-cap buffer stays out);
- closed-to-open parity for the depth gate (`a/b/def.R` under `max-depth = 1` stays out);
- boundary inclusivity (a file exactly at `max-depth` stays eligible);
- edits crossing the size boundary both ways, including clearing of the ejected buffer's own stale diagnostics;
- config change restoring eligibility (raising the cap re-admits without reopen);
- config change removing eligibility (lowering the cap ejects and clears both sides);
- ineligible open buffer keeps inlay hints while carrying no project diagnostics.

Plus two unit tests in `ry-workspace` pinning the shared policy's boundaries (inclusive size cap, parent-at-`max_depth` depth bound, root-relative counting). Local gates green: `cargo test -p ry-lsp -p ry-workspace`, `cargo fmt --all -- --check`, `cargo clippy -p ry-lsp -p ry-workspace --all-targets -- -D warnings`.

## Note on #519

This branch was rebased onto post-#519 main, so eligibility-loss clearing rides the reconciliation machinery merged in d42a1ee4 rather than inventing a parallel mechanism: an ejected buffer leaves `doc_versions` (ineligible) and its disk twin is shadowed, so it drops out of the check, and `clear_dropped_diagnostics`' existing dropped-condition (`!eligibility_for_path`) clears its tracked URI, since `eligibility_for_path` now evaluates the caps. No changes to the #519 seam itself were needed.

Closes #488

## Summary by CodeRabbit

- **Bug Fixes**
  - Open documents now consistently respect file-size, directory-depth, and exclusion limits when contributing to project-wide diagnostics.
  - Changes to document contents or configuration correctly remove or restore project participation as eligibility changes.
  - Stale diagnostics are cleared when an open document becomes ineligible.
  - Ineligible documents continue to support single-document features such as inlay hints.
  - Open buffers now take precedence over outdated indexed versions of the same file.


sims1253:
Verified and fixed: `is_file_eligible` indeed lost its last production caller to `is_file_eligible_with_limits` (only its own unit tests referenced it). Removed the wrapper in b9f1ccb3 and folded both behaviors — rooted matching and symlink entry-name matching — into the shared policy's tests. The internal `is_file_eligible_with_excludes` stays: the walker and the shared policy both still call it.

sims1253:
Applied in d7f70927: the enable gate now binds `folder_context_for_path(&path)` once and reuses it. Gates green locally (test, fmt, clippy); CI re-running on the new head.

PR #522, opened 2026-09-17
Fix stale disk index on R source changes and close

### LINKED ISSUE #486: lsp: R source changes never refresh the disk index; closing a saved file restores the last indexed snapshot

sims1253:
## Summary

Two gaps combined into #486, both verified on main with the new tests failing:

- **Unopened R sources never refresh the disk index**: `refresh_watchers` registered globs only for `ry.toml`, `DESCRIPTION`, `NAMESPACE`, native sources, and serialized data — no R source pattern — and `did_change_watched_files` early-returned for anything else. Editing, creating, or deleting an unopened `.R` file left cross-file diagnostics stale until some unrelated event triggered a full reindex.
- **Closing a saved file restores the last indexed snapshot**: `did_close` removed the doc state but never re-read the file, so after edit-save-close (with another doc open) the closed file was re-analyzed from its initialize-time content.

What changed:

- `refresh_watchers` registers `**/*.{R,r,S,s,q}` alongside the existing globs (extension set mirrors `ry_workspace` discovery); clients without dynamic registration keep today's behavior.
- `did_change_watched_files` routes R source events through a per-file refresh: changed/created files are re-read via the bounded decoder, deleted/unreadable/inadmissible files leave the index, then open documents republish. Resolution-changing events still take the full reindex path.
- New `Backend::refresh_disk_entry` applies the walk's admission rules per file — extension, symlink, pruned ancestor directories (`revdep`/`src` only as direct package children, `tests/testthat/_snaps` only package-relative), testthat runner classification, and the shared `is_file_eligible_with_limits` policy (excludes plus size/depth caps, #488) — so a rescan cannot disagree about membership. `.Rbuildignore` and directory-name excludes converge on the next rescan (documented in code).
- `did_close` re-reads the closed path through the same refresh, so save-then-close analyzes the saved bytes and discard-then-close the untouched disk bytes — no content comparison needed since the decoder is a pure function of disk state. Dropped entries reconcile through the #489 dropped-URI path on the next republish.
- Generation discipline both ways: R-only watcher batches claim a new index generation only when a refresh actually landed (saves of open docs retire nothing); when the retired pass was the initial index, a fresh pass takes over clearing `initial_index_pending` so publications never strand. The refresh skips live buffers and loses to a concurrent `did_open`.
- `ry_workspace` exports `is_r_source_path` / `is_test_fixture_path` so the LSP classifies single files with the walk's own predicates.
- The state-machine harness now sends each file's own watched event (created/deleted/renamed) alongside the `ry.toml` touch instead of faking config-only events; its model still never saves buffers, so save-then-close stays covered by the dedicated protocol tests.
- The publish path still performs zero disk I/O: all freshness work happens in the watcher/close handlers.

## Evidence

- New `crates/ry-lsp/tests/disk_index_freshness.rs` (7 tests): the exact `**/*.{R,r,S,s,q}` registration glob is asserted; unopened edit/create/delete flip cross-file RY040 shadowing live; save-then-close and discard-then-close each converge with a fresh server on the final filesystem; a seam-based test pins that a watched event retiring the initial index still leaves diagnostics flowing. Verified 0/6 pass on unfixed code (pre-rebase base), 7/7 pass fixed.
- `pr_session_seeds` passes with the harness now exercising the real `.R` event path.

## Test plan

Local gates: `cargo test -p ry-lsp` green (all 21 suites), `cargo test -p ry-workspace` green, `cargo fmt --all -- --check` clean, `cargo clippy -p ry-lsp --all-targets -- -D warnings` and `cargo clippy -p ry-workspace --all-targets -- -D warnings` clean.

Deliberate scope: no `did_save` handler (the watched-change event on save plus the close-time re-read cover both orderings); `.Rbuildignore`-relative and directory-name-exclude admission converge on rescan rather than per event; clients failing dynamic registration fall back to prior behavior. Rebased onto #519 (reconciliation) and #520 (shared eligibility) — the refresh feeds both.

Closes #486


sims1253:
Review follow-up addressed in b9d1c29f (verified against 366a450, which the incremental review covered):

- **Exact-glob assertion**: `watcher_registration_includes_r_sources` now asserts the exact `**/*.{R,r,S,s,q}` glob (disk_index_freshness.rs). The loose match the review flagged was a rebase artifact — the exact assertion existed pre-rebase and is restored.
- **Seam-based retirement test**: `watched_event_during_initial_index_keeps_diagnostics_flowing` pins the fixed behavior end to end (pause the initial pass, deliver a watched change for an unopened file, open the observer, release; diagnostics must flow and converge). Same story: present pre-rebase, dropped by the conflict resolution, now restored. 7/7 green locally with fmt/clippy clean.

No production-code changes in this commit; the handler fix the review verified (bump-only-after-refresh + respawn when still pending) is untouched.

PR #554, opened 2026-09-20
fix(lsp): make open-buffer authority path-local

sims1253:
## Problem

In a non-package workspace where `a.R` and `z.R` both define `f`, opening `a.R` with byte-identical contents changed which `f` won analysis. Whether a buffer was open, not what it contained, decided precedence: `publish_diagnostics` built one sorted list of disk files and appended a second sorted list of open files, so every open file outranked every closed file.

## Change

`publish_diagnostics` now assembles one path-keyed source view: eligible disk entries minus the disk twin of each open path, eligible open buffers inserted at their own path, sorted once by the canonical path order the corresponding CLI project uses. `reorder_files` invalidates multi-file names when the canonical order actually moves, so a warm `Project` matches a freshly assembled one after an order-only change.

This deliberately replaces the old "open files layered last, open files win" contract. The unreleased changelog entry states the new policy.

## Tests

The cross-file shadowing battery is rewritten to the new contract. Every open/closed subset of a three-file workspace now matches the fresh CLI result, and different open orders agree. Reverting the backend change fails 3 of the 8 new tests.

Evidence, red baselines, and the reproduction logs: `~/Documents/ry-hardening/evidence/p2-p1/`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed inconsistent diagnostics when multiple files define the same name.
  - Open files now replace disk contents only for their own path, while closed files retain deterministic path-based precedence.
  - Improved consistency when opening, closing, saving, or discarding file changes.
  - Warm project analysis now refreshes affected results when file ordering changes.

- **Tests**
  - Added coverage for cross-file definitions, file lifecycle changes, and consistency with fresh project checks.

---


sims1253:
All five review points are addressed in one commit:

- `cli_diagnostics` runs the CLI subprocess via `tokio::task::spawn_blocking` (owned root path), so the child wait no longer stalls the server task on the single-threaded runtime.
- `answer_watcher_registration` now panics when the `client/registerCapability` request never arrives or its response fails, instead of burning the timeout silently.
- The subset test runs the fresh-project CLI oracle explicitly before the loop instead of hardcoding "no RY040".
- The save-close test drains the watched refresh (`sync_barrier`) before taking `close_mark`, so the mark cannot be satisfied by a pre-close publication.
- The dead `let _ = scratch;` binding is removed.
---

sims1253:
All six items from the second review round are in (b4db86da, 12c5c0f4):

- `ry_binary()` now resolves inside the blocking task; its first call runs a cargo build, which is as blocking as the child wait.
- A non-JSON CLI stdout fails with the stdout and stderr heads in the panic instead of a raw serde error.
- The double-open test awaits one publication per state transition (mark after opening use.R, separate mark before opening a.R).
- The save-close drain is now a proper publication drain: a test-local `drain_publications` consumes queued pre-close publishes until an idle window passes (same end-of-stream idiom as `quiesce_diagnostics`; a deduplicated no-op refresh just ends the first window — a naive `quiesce_diagnostics` call times out on it, which is how the suppressed publication was confirmed).
- `assert_matches_cold` returns its warm run's diagnostics; the reorder test asserts the flip from that result instead of a cache-served second `check_incremental`.
- The subset doc now says "every non-empty subset" with the empty subset noted as covered separately.
---

sims1253:
All five items from the third review round, in one commit:

- `drain_publications` is bounded at sixteen consumed publications and fails with the last received message instead of hanging.
- The `same_path_buffer` and `discard_close` tests take one publication mark per transition (each open's publication awaited before the edit's own mark), so a debounce split between open and change can no longer satisfy the await with the pre-change state.
- The subset tests now assert use.R publishes no diagnostics at all, matching the asserted-empty CLI oracle — full parity with the doc's claim, not just RY040 absence.

---

sims1253:
All four items from the fourth review round:

- The save-close test takes one publication mark per transition now (its single mark spanning both opens and the change was the last such window).
- The drain cap is a named, documented constant (`DRAIN_PUBLICATION_CAP = 16`, sized to the three-file broadcast with margin).
- A no-op reorder test pins the steady-state short-circuit: same order in, warm incremental state unchanged, results identical to a fresh project both times.

---

sims1253:
Round 5: two fixed, two declined with reasons.

Fixed (e38af8a2):
- Each awaited transition is now followed by the bounded idle-window drain, so additional debounced publications left queued by the await-and-discard pattern can no longer satisfy the next mark.
- `reorder_files` documents the duplicate-path contract (last occurrence wins; callers pass the deduplicated canonical order).

Declined:
- Rebuilding definer counts per effective reorder (perf·low): the rebuild is scoped to multi-file names and runs only when the canonical order actually moved — a cold path by construction. Caching it would add invalidation bookkeeping to every file edit to save work that happens on reorders only.
- Per-name order-precision invalidation (perf·low): invalidating every multi-file name when the order moves is the deliberate conservative bound — computing per-name relative-order stability would couple this to the refinement cache's internals for a recompute saving on an already-scoped path. Correctness is unaffected either way.

---

PR #555, opened 2026-09-20
fix(lsp): reconcile watched updates until settled

sims1253:
## Problem

`refresh_disk_entry` retried a lost read twice, then fell back to a full scan. When that fallback was itself superseded, the function returned without retaining the original path. A file's update could then be forgotten with no further event coming to rescue it.

## Change

Every refresh enqueues a per-path obligation at its epoch claim, before the read that might lose. Terminal outcomes settle exactly the revision they cover; every ladder loss, including a superseded `SettledWithoutInstall` backstop scan, retains the obligation. One reconciliation driver re-runs the handler pipeline for pending paths. The driver never preempts an in-flight refresh, goes idle under the lock when everything is in flight, and is stall-bounded: eight no-progress rounds drop an obligation with a `tracing::warn`. Folder removal and shutdown purge obligations.

Stacked on the P2 branch (base: hardening/p2) so the ordering policy lands first.

## Tests

The gate-driven schedule (first read loses to B, retry to C, fallback scan to D, then silence) was red before and is green after, in both directions: error becoming clean and clean becoming error. Removing the reconciliation step fails its tests. A happy-path watched event still spawns zero driver rounds.

One known sensitivity: the zero-spawn overhead assertion is timing-sensitive and failed once in eight runs; convergence itself was never violated.

Evidence: `~/Documents/ry-hardening/evidence/p2-p1/`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed language-server file refreshes so watched-file and close-time changes reliably converge to current diagnostics, even during concurrent updates.
  * Prevented stale diagnostics from remaining after refresh conflicts or bursts of file events.
  * Improved shutdown and workspace-removal handling to cancel pending refresh work cleanly.
  * Added safeguards against stalled refresh processing and unnecessary repeated work.


sims1253:
Review follow-up (3884a69a). Per-item:

1. Stall detector measured progress as a shrink of `pending_refreshes` — confirmed; arrivals offset retirements. Progress is now a monotonic retirement counter (incremented where an entry is removed by `settle_pending_refresh`/`complete_pending_publication`), snapshotted per round; cancellations (folder removal, shutdown, the stall bound) remove without retiring and do not count as driver progress. `RECONCILE_MAX_STALLED_ROUNDS`'s doc, the state-field doc, and the CHANGELOG entry now say exactly that. Pinned by a unit test driving eight offsetting rounds (retire one, arrive one — map size constant) with no drop, then eight retirement-free rounds dropping the survivor; a full eight-round interleaving is not deterministic in the gate harness, so the accounting is unit-tested directly.

2. Lost wake at the idle-out — confirmed; the duty checks ran under earlier, separate lock holds. The idle branch now re-derives its verdict (`has_drivable_obligation`) under the same lock hold that clears `reconcile_driver_active`, and continues driving instead of clearing when something became drivable in between. New test `driver_idle_out_keeps_the_signal_of_a_refresh_exiting_in_the_window` parks the driver between its last duty check and the flag clear via a new driver-idle gate, releases the losing refresh inside that window, and asserts convergence; removing the re-check makes it fail deterministically.

3. Shutdown race — confirmed; the purge only helped a driver between rounds. `shutdown` now sets `shutting_down` and purges in one critical section; the driver re-checks the flag under the lock at the round head, at every in-round dispatch, and before the follow-up publications, and `claim_refresh_epoch` refuses post-shutdown claims so no event can re-seed the emptied map. New test `shutdown_during_a_driver_round_publishes_nothing_afterwards` parks the driver's own dispatched refresh at the commit gate, answers a raw shutdown request (no `exit` — sending exit closes the socket and would suppress any publication regardless of the fix), then asserts a bounded negative window stays empty. Dropping just the flag makes the test fail with the leaked publication; existing shutdown/teardown tests stay green.

4. `note_refresh_exit`'s bool is now used: an exit that leaves an idle obligation still owing its bytes phase wakes the driver from the exit itself; a landed exit leaves the context/publication follow-up to its dispatcher and does not wake (waking there would spawn a driver on every unopposed save and break coalescing — the zero-spawn happy path still observes zero spawns). One wrinkle the old "no wake here" comment gestured at: waking from the ladder plus the inline `tokio::spawn` in `wake_reconciliation` makes the driver future's Send obligations circular; the spawn moved behind a plain fn boundary (`spawn_reconciliation_driver`), which breaks the cycle.

5. Undercount confirmed: the root-removal retain (also the stall-drop and shutdown clears) could drop an entry whose refresh was still in flight, and a later event's re-armed obligation would absorb the old exit's decrement — the driver could then preempt an in-flight read. In-flight accounting moved to a side map (`refreshes_in_flight`) written only by the claim and exit critical sections, independent of the obligation map's removals. Unit test: cancel + re-claim + old exit must not report the obligation drivable; only the owning exit may.

6. Both style items: `RefreshDuty`/`PendingRefresh` are now module-private (nothing outside backend.rs names them), and the stall warning's stray-space run is joined.

Also in this push, the CI flake in `event_burst_coalesces_and_converges` (run 35522769680): the zero-spawn assertion is no longer post-hoc process-global. It is delta-scoped from a mark taken after `settled_ladder` now awaits the close handler's empty-clear publication (proving the context settle, obligation completion, and clear finished before the event under test), and the event's landing is proven through the post-commit gate rendezvous. Zero spawns is structural at that point: a landed exit leaves only dispatcher-owned duty (the exit-side wake skips it by duty), and the handler completes the obligation before its epilogue wake, so the wake finds an empty map whenever it runs. Battery re-run under load: 10/10 with full logs (6 spinning cores) plus 5/5 alongside a concurrent workspace build; one earlier scripted check-run reported a failure with its output discarded and passed on the immediate rerun — the 15 logged runs since were clean. `watched_closed_publish`'s ladder test is untouched by the restructure (its gates are unchanged) and stays green.

Gates: cargo test -p ry-lsp (all binaries, incl. the eight focused ones), cargo test --workspace, clippy --workspace --all-targets -D warnings, fmt --check: all green.
---

PR #557, opened 2026-09-20
fix(cli): watch nested package metadata dependencies

sims1253:
## Problem

Watch mode collected metadata candidates only from explicit roots and their ancestors. Editing a nested package's `NAMESPACE`, or creating a nested `DESCRIPTION`, changed nothing until a source file also changed, so watch output diverged from a fresh one-shot check.

## Change

`package_meta_paths` now derives candidates from every discovered R file's ancestor chain up to the first covered ancestor, and watches `DESCRIPTION` candidates even while absent, so creating a nearer package root is observable. The stamp set follows membership and discovery-config changes; newly added dependency paths force a check instead of seeding current timestamps. No recursive scan was added; the cost is two `stat` calls per unique ancestor directory per poll.

## Tests

Nested NAMESPACE edit, create, and delete; nested DESCRIPTION create and delete; an initially empty workspace gaining a package and source; config changes admitting excluded sources; and sibling packages all converge with a fresh check. An unrelated-metadata control outside the dependency set stays quiet. Reverting the fix fails exactly the six nested-metadata witnesses.

Evidence: `~/Documents/ry-hardening/evidence/p4/`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved watch mode tracking for package metadata, including nested packages and explicitly selected files.
  * Changes to relevant DESCRIPTION and NAMESPACE files now trigger re-checks when package relationships or scopes change.
  * Improved handling of relative paths and packages added to previously empty workspaces.
  * Prevented unrelated package trees and metadata outside the applicable package boundary from triggering unnecessary re-checks.


sims1253:
Review follow-up on the metadata-dependency watch set, item by item. Commits: ce9f6d80 (src), b146396c (tests).

**1. Unbounded root chains (bug, medium) — fixed.** Chains now stop at the first ABSORBING ancestor: a directory already covered by an earlier chain, a directory holding an existing DESCRIPTION (a package boundary; a fresh one-shot groups files by the nearest existing DESCRIPTION and stops there, so nothing above it can change resolution), or the config root (the `ry.toml` anchor and the fallback resolution root). Absorbing ancestors are included before stopping; with no absorber anywhere the chain still runs to the top, exactly as far as the one-shot grouping walk. Verified red on the previous head: for a root under /tmp the set contained `/DESCRIPTION`, `/NAMESPACE`, `/tmp/DESCRIPTION`, and creating DESCRIPTION/NAMESPACE in the parent of a package-root watch root triggered a spurious re-check. Now pinned by the unit test `root_chain_stops_at_first_existing_description_ancestor`, `root_chain_stops_at_the_config_anchor`, and the e2e `watch_ignores_metadata_above_absorbing_boundary` (silent above the boundary, still reactive to the boundary's own NAMESPACE edit). One documented trade in the `package_meta_paths` doc: a package root created above an absorbing ancestor is not watched, though a fresh check would honor it — the price of keeping /, /tmp, $HOME out of the set. Deleting an absorbing DESCRIPTION is watched, and the poll after any observed difference re-derives the set, so a moved boundary self-heals.

**2. Relative inputs terminate early (bug, low) — fixed.** Roots, discovered paths, and the anchor are resolved through a new `absolutize` helper (working-directory join plus lexical normalization) before walking; `canonicalize` is not an option because watched candidates may be absent by design. Verified red on the previous head: a relative root produced the bare candidates `DESCRIPTION`/`NAMESPACE` next to `pkg/DESCRIPTION` and the walk stopped at the working directory instead of rising past it (also hit `ry check . --watch`, which watched `./DESCRIPTION` and `DESCRIPTION` — the same file twice). Pinned by `relative_roots_walk_absolute_ancestors` (unit) and `watch_relative_root_reacts_to_package_metadata` (e2e: child spawned with the fixture as cwd and `pkg/R/use.R` as input). `config_candidates` now reuses the helper.

**3. Per-poll re-derivation (performance, medium) — fixed.** `WatchState` caches the derived candidate set; a poll re-derives only when the file set changed, the discovery config reloaded, or an observed stamp difference may have moved an absorbing boundary. Idle polls stat the cached members and nothing else. Re-measured on the same fixture shape as the PR report (63 files, 6 nested packages, 2 R-less sibling metadata trees, quiet 11 s window, debug build, two runs each): previous head 0.090/0.092 s user, fixed 0.082/0.084 s. On that fixture the candidate set is identical before and after (the six package DESCRIPTIONs stop the file chains either way), so the delta is purely the removed derivation; the +2-stats-per-ancestor cost now applies only to ancestors below the first absorbing boundary.

**4. Cross-pipe ordering race (test, medium) — fixed.** The exact stdout occurrence-count assertion in `watch_empty_workspace_gains_package_and_metadata` now waits for the stdout finding itself first (`wait_for_more` on the finding text — the sibling-package test's existing pattern), since stdout and stderr are separate pipes drained by independent threads. No sleeps; the other count assertions in the suite already synchronize on their predecessor's stdout.

**5. Closure captures `seen` but takes `paths` (maintainability, low) — fixed.** The collector now captures both and takes neither; it also checks `contains` before inserting, so the already-covered common case allocates nothing.

**6. Snapshot order stability (maintainability, low) — declined with reason.** `package_meta_paths` already returns `paths.sort(); paths.dedup();`, so snapshot order is deterministic regardless of the order `discovered` arrives in (`rescan` sorts as well). The doc comment on `snapshot_meta_paths` that claimed insertion-order stability "without sorting" was inaccurate and is rewritten to point at the explicit sort; a sortedness assertion is pinned in the existing bounded-set unit test.

Gates on the final tree: `cargo test -p ry-cli` (41 unit tests incl. 3 new; watch_e2e 18/18, run three times), `cargo test --workspace` (73 suites, zero failures), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all clean.
---

sims1253:
Second review round — all five items addressed (aa2de68b, f1b69316, 9f0dddea).

**1. CodeRabbit Major, check.rs:685 — `config_candidates` vs `Config::discover` normalization.** Verified real by construction before fixing: a unit test placed the only `ry.toml` in `<base>/sub` and ran both walks from `<base>/sub/../project`. Discovery selected `<base>/sub/ry.toml` (its `Path::parent` walk steps over the dots, visiting `<cwd>/..` and `<cwd>` itself), while the folded candidate list was `[<base>/project/ry.toml, <base>/ry.toml, /tmp/ry.toml, /ry.toml]` — the active config absent from the watch set, so `poll_static_inputs` would never see edits to it. The failing-first run is captured in the branch evidence. Fix: split discovery's join (no `..` folding) out as `discover_start` and anchor `config_candidates` on it, so both walks visit the same directories in the same order; `absolutize` keeps the folding for the package-metadata chains only. ry-config's public behavior is untouched. The test (`config_candidates_mirror_discovery_for_dotdot_inputs`) pins dotted directory and dotted file inputs: the first existing candidate must equal discovery's pick.

**2. OpenCodeReview test·medium, watch_e2e.rs:775 — exact-count race.** `watch_sibling_package_metadata_edits_stay_scoped` now waits for the counted stdout content itself (`wait_for_more` on `` `missing_in_b` is not bound ``) before the exact occurrence assertion, same pattern as the sibling fix.

**3. CodeRabbit Minor, watch_e2e.rs:1029 — baseline captured after the event.** Both negative metadata tests (`watch_ignores_metadata_above_absorbing_boundary`, `watch_ignores_unrelated_package_metadata`) now capture `passes_before` before the metadata writes, so a regression that does re-check on that metadata can no longer slip the extra pass into the baseline and pass unnoticed.

**4. OpenCodeReview maintainability·low, check.rs:732 — symlink/`..` folding.** Documented the limitation precisely on `absolutize` rather than hardening: a `..` crossing a symlinked component folds to a path the OS never resolves to that file, so the residual risk is a missed auto-refresh on such trees — never a spurious pass, never a wrong result (the pass that runs reads what a fresh check would; the next unrelated edit or a restart picks the change up). Canonicalizing the deepest existing ancestor was rejected: it would stat up the tree on every derivation and churn snapshot entries as creations make deeper prefixes canonicalizable, and the one-shot walks (`Path::ancestors`, `Config::discover`) are lexical by design too.

**5. pullfrog, check.rs:764 — over-broad trade-off claim.** Confirmed against `enclosing_package_root` (lexical `ancestors()`, stops at the first existing DESCRIPTION, no config-root stop): the divergence exists only above the CONFIG ROOT absorber. The comment now claims the trade for the config root only and states that above a DESCRIPTION boundary the stop loses nothing — the nearest-boundary rule hides a higher root from a fresh check as well.

Gates: `cargo test -p ry-cli` (watch_e2e run twice, 18/18 both), `cargo test --workspace`, `clippy --workspace --all-targets -- -D warnings`, `fmt --all -- --check` — all clean.

---

sims1253:
Round 3 (OCR 35534990805) — all nine items triaged against `9f0dddea` and addressed; nothing declined. Landed as `d9368e23..32ab0b47`.

**watch_e2e.rs**

- **test·medium :785 (exact-count across two pipes)** — the sibling-scope test (and the same-shaped phase of `watch_empty_workspace_gains_package_and_metadata`) now synchronizes in the order that guarantees the counted lines are complete: wait for this pass's stdout lines first, then for the next stderr summary past them — the pass writes its summary last, so the summary's arrival proves the child finished the pass's output writes, and the buffered stdout waits cover pump lag — and only then read the count. (32ab0b47)
- **test·medium :1000 / test·low :1052 (fixed 2500 ms negative window; shared constant)** — both negative tests now share one `NEGATIVE_WINDOW` constant (five polls at the watch loop's 500 ms interval, rationale next to the phase-discipline assumption). The window is bounded by the preferred positive-control pattern: `watch_ignores_unrelated_package_metadata` gained the trailing control the absorbing-boundary test already had (a watched-metadata edit that must re-check), so a merely slow watcher still fires the control — proving the quiet was genuine — while a stalled one fails for the right reason instead of passing vacuously. Both tests also drain the initial pass summary before baselining the count, removing the one real loaded-runner false-failure vector (a summary landing after the "watching" line). The window stays, so the constant item did not disappear. (32ab0b47)
- **doc·low :490 (module comment over-claims "no sleeps")** — module header and the nested-package section comment now scope the claim precisely: positive assertions are deadline-based content waits; fixed sleeps appear only as the early-exit grace period and the control-bounded `NEGATIVE_WINDOW`. (32ab0b47)

**check.rs**

- **perf·medium :575 (double derivation when `rederive`)** — `poll_meta_inputs`' `rederive` path now derives the set and its snapshot exactly once per poll (via the new `WatchState::derive_meta_state`) and installs that fresh pair directly on a difference; the just-run derivation already reflects any absorbing-boundary move, so the old unconditional tail call rebuilt an identical set. Only the cached path keeps the post-difference re-derivation, which is the only path that can have missed a boundary move. Verified by derivation count on a watch session growing the file set 12 times (each growth poll is a `rederive`+difference poll): 25 derivations before, 13 after — the redundant second derivation per differing poll is gone; the settle semantics (pre-check stamps installed, one extra settle pass after a boundary move) are unchanged. (d9368e23)
- **maint·low :520 (derive-and-snapshot helper)** — `derive_meta_state` is now the single derive/snapshot pairing, shared by `sync_meta_stamps` and the poll's `rederive` path; the cached path's deliberately unpaired tail derivation keeps its cross-reference and invariant note in the method docs. (d9368e23)
- **maint·low :723 (`discover_start` drift)** — declined the shared `ry_config` helper and documented why on the function: the anchor is three lines with no behavior of its own and a single caller outside the crate, so a shared home would widen ry-config's public surface and couple the crates for a walk this small; `config_candidates_mirror_discovery_for_dotdot_inputs` is named as the drift guard that fails when discovery's anchoring changes (re-verified the copy is still verbatim first). (e29409c8)
- **doc·low :739 (cwd-failure fallback divergence)** — `absolutize`'s accepted-limitation note now covers both divergence cases: symlink-crossing `..`, and the `current_dir()`-failure `.` fallback folding a relative input's leading `..` against an empty stack — each missed-auto-refresh at worst, healed by the next edit or restart. (e29409c8)
- **test·low :1640 (depth-dependent ancestor test)** — the bounded test's watched site now sits under a constructed fixed-height ladder (`tower/a/b/site`) whose top `DESCRIPTION` absorbs the climb, so the expected set is exactly the fixture's (9 dirs / 18 candidates) no matter the checkout depth or sandbox layout; a depth-independent one-level-up pin in the explicit-file-root test keeps the no-absorber-climbs-past-root behavior covered. (210a1c2e)

Gates: `cargo test -p ry-cli` green (bin unit 42, watch_e2e 18/18 — run three times on the final semantics), `cargo test --workspace` 73 suites green, `clippy --workspace --all-targets -- -D warnings` clean, `fmt --all -- --check` clean.

---

PR #559, opened 2026-09-20
fix(checker): respect NA and dispatch in RY107 claims

sims1253:
## Problem

RY107 classified `any()`/`all()` comparisons as FALSE = 0 and TRUE = 1 and worded the result unconditionally. `any(NA) > 1` is NA, not FALSE, and a direct or Summary method, or an S4 method behind `base::any`, can return a value outside that range.

## Change

Constant-result wording now says the result holds "when the base result is not NA (an NA result compares as NA)". The heuristic stays framed as a suspected element-level comparison. The rule keeps firing on open-world arguments; authors of S4 methods still get the base-domain warning. That tradeoff preserves the founding glue true positive.

## Tests

Runtime witnesses for NA outcomes and for direct and Summary dispatch are asserted in R inside the oracle fixture. The glue vendor snapshot keeps its single triaged true positive, the diffobj quiet control stays quiet, and RY105, RY093, and RY100 behavior is pinned unchanged.

Evidence: `~/Documents/ry-hardening/evidence/p3-p5-p6/p6/`.

## Summary by CodeRabbit

- **Documentation**
  - Clarified RY107 diagnostic messages to account for `NA` results from `any()` and `all()`.
  - Documented that method dispatch can change the expected result domain.
  - Updated the rule registry, rule documentation, changelog, and example diagnostic.

- **Tests**
  - Added coverage for `NA` propagation, negating comparisons, and dispatched methods.
  - Confirmed existing RY107 detections and firing conditions remain unchanged.

---


sims1253:
Addressed this cycle's feedback:

- Confirmed the S3 point by execution: a local `Summary.s3grp <- function(x, ...) 42` intercepts `any()` (base also ships `Summary.data.frame`). 'Unless an S4 method dispatches' under-covered the premise, so the message, docs, and test comment now name `any`/`all` and `Summary`-group methods, and the oracle claim fixture executes the S3 group witness instead of leaving dispatch to prose (bf6566c2).
- The fourfold NA/S4 justification is trimmed to one statement in the rule; module/function docs keep the full story.
- The `posit` failure was the expected message-ledger drift (glue RY107 wording). Regenerated `posit-messages-0.9.json`; the identity and all reports are unchanged (7a324844).
---

sims1253:
Three follow-ups from this round, in one commit:

- The registry summary and CHANGELOG entry still said "an S4 method (direct or `Summary` group)" after the broadened message; both now match the shipped wording (an S4 `any`/`all` method, or an S3/S4 `Summary` group method).
- The premise string duplicated across the two `format!` branches is hoisted to one `premise` binding; message text is byte-identical (snapshot and ledger untouched, readme_rule_table green).
---

PR #560, opened 2026-09-20
chore(typeshed): integrate pinned release candidate

sims1253:
## Change

Vendors r-typeshed `cd81a0f` via `scripts/sync_typeshed.sh`; `vendor/SOURCE` and the stubs digest are recorded. Registers MASS, Matrix, RColorBrewer, curl, gridExtra, xml2, and zoo in `packages.txt` in sorted position, preserving the order of all existing entries.

## Verification

With the pinned consumer (`b97cc653`): `typeshed validate` passes with 55 stubs and 0 errors, the inference gate passes 19/19, and all three mutation self-tests (density-length, as-vector-arg0, append-concat) fail for the intended semantic mismatches. Embedded and `--typeshed`-override loading agree on the new inventory: 14/14 check fixtures and 5/5 dump fixtures identical. Seven new loader tests cover qualified and namespace-import resolution for a representative function and exported value per package.

Both local overlays stay: the candidate ships no `vec_cast`, and base's `median` still lacks the verified `na.rm` formals.

## Notes

The embedded version-pin test followed the sync (0.0.19 to 0.0.26). One sapply test was rewritten because the candidate types the sqrt/math family as `union(double, complex)`, so the simplify rule now declines on that input. That is correct semantics and a true-positive loss on non-negative input, recorded here rather than hidden.

Evidence: `~/Documents/ry-hardening/evidence/p7-p8/p7/`.

---


sims1253:
Both stub-data items from review, triaged:

- **Matrix `crossprod`/`tcrossprod` `params: []` — declining, with evidence.** The entry mirrors what R exposes: `Matrix::crossprod` is the base `.Primitive`, `formals()` of a primitive is NULL, and r-typeshed's own audit (`tests/Matrix.R`) enforces `params == names(formals(getExportedValue(...)))` — I tried `["x","y","..."]` (the documented signature) and the audit fails on exactly that identity. Behaviorally there is no false arity-mismatch: `Matrix::crossprod(m)` and `Matrix::tcrossprod(m)` check with zero diagnostics through the embedded loader (`RY_NO_INSTALLED_LIBRARIES=1`, release binary) — empty `params` is read as no arity claim, not zero-arity. If you want primitives' documented formals in stubs, that is an r-typeshed audit-policy change (derive from `args()` for primitives) and belongs there, not in this sync PR; say the word and I'll prepare it.
- **zoo.json trailing space — fixed upstream** in sims1253/r-typeshed#92 (one byte; validate 55/0, `tests/zoo.R` green). The vendored copy predates it by design; the next sync picks it up, so this PR keeps vendor exactly at cd81a0f.
---

